Issue #61, opened 2019-10-10
Fill CustomField interface with the fields


rmakara:
**Context**
Right now CustomField interface is empty. We want to fill it based on the knowledge from Shopware 6 sources.

**Acceptance criteria**
* `CustomField` interface contains the same fields as original CustomField in Shopware 6.

rmakara:
@mkucmus 
Is it done in #107  and #110  ? :)

mkucmus:
CustomField interface is not being used. Can be closed.

Issue #43, opened 2019-10-04
Prepare TypeScript interfaces for Shopware models, that we will receive from Shopware API


rmakara:
**Context**
Before we prepare API client and methods for all the requests required in shopware-pwa, we would like to prepare Shopware models to make it much easier to operate the data structures that we will receive in responses. 

In that task we want to prepare models required to get responses from `sales-channel-api` and its `category`, `product`, `customer`, `cart`, `context`, `newsletter` endpoint groups. For building the interfaces we may use our Postman for Shopware API (#3), symfony debug router and Shopware source code itself.

Shopware provides following structure of the root folder, that we should stick to:
* checkout,
* content,
* framework,
* system.

Right now, there is no need to copy the structures in 100%. Just the ones that seems to be reasonable to operate on the API. Within' the next tasks, when we implement API endpoints we will be able to adjust the models structure.

**Acceptance criteria:**
* All the models that seem to be required for operating on `sales-channel-api` endpoints exist in shopware-pwa.
* `models` directory exist in a tree and it contains `checkout`, `content`, `framework`, `system` subfolders.
* 1 level of nested folders below `checkout`, `content`, `framework`, `system` has been recreated bases on Shopware original structure. Levels below are not required at this moment.

PR #65, opened 2019-10-10
#43 - Add interfaces for common models of Shopware 6

rmakara:
**Changes:**
* Add interfaces for `common` group.
* Right now it contains following interfaces: `Entity`, `Collection`, `CustomField`.

**To do:**
* `CustomField` interface to be filled in #61 

**Issue:**
#43

### DIFF
--- a/packages/shopware-6-client/src/interfaces/models/common/Collection.ts
+++ b/packages/shopware-6-client/src/interfaces/models/common/Collection.ts
@@ -1,3 +1,0 @@
-export interface Collection {
-  elements: [];
-}
--- a/packages/shopware-6-client/src/interfaces/models/common/Collection.ts
+++ b/packages/shopware-6-client/src/interfaces/models/common/Collection.ts
@@ -1,0 +1,3 @@
+export interface Collection {
+  elements: [];
+}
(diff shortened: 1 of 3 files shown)


packages/shopware-6-client/src/interfaces/models/common/Entity.ts
```diff
@@ -0,0 +1,7 @@
+export interface Entity {
+  _uniqueIdentifier: string;
+  versionId: string;
+  translated: [];
```
patzick: Which type of Array do we have here?

packages/shopware-6-client/src/interfaces/models/common/Collection.ts
```diff
@@ -0,0 +1,3 @@
+export interface Collection {
+  elements: [];
```
patzick: please provide an example of this collection usage

also, we need to add a type of elements array

packages/shopware-6-client/src/interfaces/models/common/Entity.ts
rmakara: Just an array in Shopware. No specific type.

packages/shopware-6-client/src/interfaces/models/common/Collection.ts
rmakara: We can delete this one as we deleted all the interfaces suffixed with Collection.

rmakara:
@patzick suggestions applied. :)

Issue #70, opened 2019-10-11
I want to prepare ENDPOINTS.TS file So everyone may use the consts for API endpoints


rmakara:
**Context**
In `packages`shopware-6-client/src/endpoints.ts` we want to store consts to all the endpoints available in Shopware 6 API. In all the other modules of the package we will be able to refer to them by using a const instead of hardcoded string.

**Acceptance criteria**
* Every endpoint is defined as const.
* All the consts are exported.
* Following endpoints are defined (or documented as JSDoc) in the file:
  * Category
    * GET /category
    * GET /category/{{categoryId}}
  * Product
    * GET /product
    * GET /product//{productId}
    * GET /product//{productId}?associations[][]
  * Customer
    * Address
      * GET /customer/address
      * GET /customer/address/{{address_id}}
      * POST /customer/address/
      * DELETE /customer/address/{{address_id}}
      * PATCH /customer/address/{{address_id}}/default-shipping
      * PATCH /customer/address/{address_id}/default-billing
    * POST /customer
    * POST /customer/login
    * GET /customer
    * GET /customer/{{customer_id}}
    * POST /customer/logout
    * GET /customer/order
    * GET /salutation
  * Cart
    * POST /checkout/cart
    * GET /checkout/cart
    * POST /checkout/cart/product/{{product_id}
    * POST /checkout/cart/line-item/{{line_item_id}}
    * DELETE /checkout/cart/line-item/{{line_item_id}}
    * POST /checkout/order
    * POST /checkout/guest-order
    * POST /checkout/order/{{order_id}}/pay
    * GET /checkout/guest-order/{{order_id}}
  * Context
    * GET /currency
    * GET /language
    * GET /country
    * GET /payment-method
    * GET /shipping-method
  * Newsletter
    * POST /newsletter/subscribe
    * POST /newsletter/unsubscribe

PR #75, opened 2019-10-14
feature/70-api-endpoints-consts

rmakara:
**Changes:**
* Added methods to `endpoints.ts` for Shopware API.
* Added tests for `endpoints.ts`.

**Issue:** #70

### DIFF
--- a/packages/shopware-6-client/__tests__/endpoints.spec.ts
+++ b/packages/shopware-6-client/__tests__/endpoints.spec.ts
@@ -1,0 +1,242 @@
+import {
+  PRODUCT_ENDPOINT,
+  PRODUCT_DETAILS_ENDPOINT,
+  CATEGORY_ENDPOINT,
+  CATEGORY_DETAILS_ENDPOINT,
+  CUSTOMER_ADDRESS_ENDPOINT,
+  CUSTOMER_ADDRESS_DETAILS_ENDPOINT,
+  CUSTOMER_ADDRESS_SET_DEFAULT_SHIPPING_ENDPOINT,
+  CUSTOMER_ADDRESS_SET_DEFAULT_BILLING_ENDPOINT,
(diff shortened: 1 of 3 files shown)


packages/shopware-6-client/__tests__/endpoints.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+const sampleCustomerId = "8b67c1fbb718487db750651430023298";
+const sampleLineItemId = "042df8812942487bb52c9fc1e5b26e20";
+const sampleOrderId = "27356105cf9b4484b96143881c37bbcb";
+
+describe("endpoints", () => {
+  describe("PRODUCT_ENDPOINT", () => {
+    it("should return Shopware product endpoint", async () => {
+      const result = await PRODUCT_ENDPOINT();
+      expect(result).toEqual("/product");
+    });
```
patzick: Endpoint methods does not return a promise :)
We shouldn't `await` for non-promise method

```ts
    it("should return Shopware product endpoint", () => {
      const result = getProductEndpoint();
      expect(result).toEqual("/product");
    });
```

packages/shopware-6-client/src/endpoints.ts
```diff
@@ -1 +1,79 @@
-export const CATEGORY_ENDPOINT = "/category";
+// category
+
+export const CATEGORY_ENDPOINT = () => `/category`;
```
patzick: yeah, in fact, the name suggests that this is a const, not method
les's change these to a more method-conventional name like `getXxxEndpoint`
```suggestion
export const getCategoryEndpoint = () => `/category`;
```

rmakara:
@patzick fixes applied

mkucmus:
Yeah! now it's clear and ready to use.

Issue #63, opened 2019-10-10
Create Customer service interface


mkucmus:
Customer service should be using a Connection layer (based on axios library).
Each method of the service will contain specific types of parameters and return typed data, described in this issue on the bottom.

_Customer Interfaces listed at the bottom of the issue._

_It's assumed that the context token is provided on request layer._

### The list of proposed methods:
```
1. login(CustomerLoginParam): CustomerResponse<CustomerLoginResponse>|401NotAuthorizedResponse
2. logout(): 204NoContentResponse
3. orderList(Pagination interface from #55 )::<SearchResult<Order[]>
4. register(CustomerRegisterParam): CustomerResponse<string|uuid> # customerId returned
5. saveEmail(CustomerChangeEmailParam): 204NoContentResponse
6. savePassword(CustomerChangePasswordParam): 204NoContentResponse
7. saveProfile(CustomerChangeProfileParam): 204NoContentResponse
8. getCustomerDetail(): CustomerResponse<Customer> # data model Customer
9. getAddresses(): CustomerResponse<CustomerAddress[]> # data model CustomerAddress
10. getAddress(addressId:string # or dedicated CustomerAddressIdParam interface should be used?): CustomerResponse<CustomerAddress>
11. createAddress(CustomerAddressParam): CustomerResponse<string | uuid>
12. deleteAddress(CustomerAddressIdParam):  204NoContentResponse
13. setDefaultShippingAddress(CustomerAddressIdParam): CustomerResponse<string | uuid>
14. setDefaultBillingAddress(CustomerAddressIdParam): CustomerResponse<string | uuid>
```

**All the parameters should be converted into query string (or payload) just before sending a request.**
Customer Params converter was described here: #TODO

### Parameters:
 Ad. 1. 
```
interface CustomerLoginParam  {
  username: string
  password: string
}
```
Ad. 2. N/A

Ad. 3. N/A

Ad. 4.
```
interface CustomerRegisterParam {
  salutationId: String|uuid,
  firstName: String,
  lastName: String,
  password: String,
  email: String
  title?: String,
  birthdayYear?: number,
  birthdayMonth?: number,
  birthdayDay?: number,
  billingAddress: BillingAddress,
  shippingAddress?: ShippingAddress
}
```
Ad. 5. 
```
interface CustomerChangeEmailParam {
  email: String
  emailConfirmation: String
  password: String
}
```
Ad. 6. 
```
interface CustomerChangePasswordParam {
  password: String
[...]

PR #77, opened 2019-10-15
feat(customer): implement Customer service interface

mkucmus:
Related issue: #63


packages/shopware-6-client/__tests__/services/customerService.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+    });
+  });
+  describe("logout", () => {
+    it("should return no data and throw no exceotion (204)", async () => {
+      try {
+        const contextTokenResult = await CustomerService.login(
+          loginCredentials
+        );
+        const result = await CustomerService.logout(
```
mkucmus: **TODO**

packages/shopware-6-client/src/services/customerService.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @description End up the session
+ */
+const logout = async function(contextToken?: string): Promise<null> {
+  const resp = await axios.post(
+    `${config.endpoint}${CUSTOMER_ENDPOINT}/logout`,
+    null,
+    {
+      headers: {
+        /** TODO: move into different layer if created */
+        "sw-context-token": contextToken
```
mkucmus: **TODO**

packages/shopware-6-client/__tests__/services/customerService.spec.ts
```diff
@@ -0,0 +1,132 @@
+import { CustomerService } from "../../src/index";
+
+const generateUniqueEmail = (): string => {
+  const currentTime = Date.now();
+  const email = `mkucmus+${currentTime}@divante.com`;
+  return email;
+};
+
+const basicCustomerData = {
```
mkucmus: @patzick do we have here a good place to apply faker generated data?

mkucmus:
I've covered all the methods which are available through the REST API. 

@patzick @rmakara I've got this feeling that I should have changed some customer methods and prefix them with "customer" word if we still want to expose all of them in one entrypoint (src/index.js)

packages/shopware-6-client/src/services/customerService.ts
```diff
@@ -0,0 +1,285 @@
...
+): Promise<CustomerLoginResponse> {
+  const resp = await apiService.post(`${getCustomerEndpoint()}/login`, params);
+  return resp.data;
+};
+
+/**
+ * @description End up the session
+ */
+export const logout = async function(contextToken?: string): Promise<null> {
+  const resp = await apiService.post(`${getCustomerEndpoint()}/logout`, null, {
```
patzick: we need to have these steps here

1. if contextToken is provided, then call `update({contextToken})`
2. call api here
3. call `update({contextToken: ''})`

don't return anything as it's a void method

packages/shopware-6-client/src/services/customerService.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+  });
+  return resp.data;
+};
+
+/**
+ * @description End up the session
+ */
+export const getCustomer = async function(
+  contextToken: string
```
patzick: we don't need to pass contextToken anywheere else than `setup` or `update`
```suggestion
```

packages/shopware-6-client/__tests__/services/CustomerService/deleteCustomerAddress.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it("returns no data if successfully deleted", async () => {
+    mockedAxios.delete.mockResolvedValue(null);
+
+    await deleteCustomerAddress("2bbb89dfa4664bc681e80b37eaa80fb7");
+    expect(mockedAxios.delete).toBeCalledTimes(1);
+    expect(mockedAxios.delete).toBeCalledWith(
+      getCustomerAddressEndpoint("2bbb89dfa4664bc681e80b37eaa80fb7")
+    );
+  });
```
rmakara: Possible to add test for following behavior:

`Customer address with id \"324af469318f46b68e0fe69d77ef15fb\" is a default address and cannot be deleted.`

However, it is not necessary for now.

packages/shopware-6-client/__tests__/services/CustomerService/getCustomerAddress.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    jest.resetAllMocks();
+
+    contextToken = random.uuid();
+
+    update({ contextToken });
+  });
+  afterEach(() => {
+    // expect(
+    //   mockedAxios.defaults.headers.common["sw-context-token"]
+    // ).toEqual(contextToken);
```
rmakara: We may delete that comment. : ) Probably it is one of the tested scenarios during remote pair programming, which is no longer needed. : )

packages/shopware-6-client/__tests__/services/CustomerService/getCustomer.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    expect(mockedAxios.get).toBeCalledTimes(1);
+    expect(mockedAxios.get).toBeCalledWith(getCustomerEndpoint());
+    expect(result.id).toEqual("c370eb5cd1df4d4dbcc78f055b693e79");
+  });
+
+  it("should an current customer's data - using incorrect token", async () => {
+    mockedAxios.get.mockRejectedValue(new Error());
+    expect(getCustomer()).rejects.toThrow();
+    expect(mockedAxios.get).toBeCalledTimes(1);
+    expect(mockedAxios.get).toBeCalledWith(getCustomerEndpoint());
```
rmakara: I am not sure if it's needed but when testing a behavior of some methods, that use incorrect token we may overwrite the correct token in config.contextToken with some invalid one. 

Right now we are testing response from the API, which is mocked to Error() but we are not sure about behavior of `getCustomer()` method. If that would be a black box method then we could assume there could be some validation, that will not even call `get` http method if token does not meet some format requirements.

Not a required change but from my perspective erasing the correct token would be a good step. :) Up to you to decide. :)

packages/shopware-6-client/__tests__/services/CustomerService/getCustomerAddresses.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    afterEach(() => {
+      expect(config.contextToken).toEqual(contextToken);
+    });
+  });
+
+  it("should return object of addresses", async () => {
+    mockedAxios.get.mockResolvedValueOnce({ data: { data: {} } });
+    const result = await getCustomerAddresses();
+    expect(mockedAxios.get).toBeCalledTimes(1);
+    expect(mockedAxios.get).toBeCalledWith(`${getCustomerEndpoint()}/address`);
```
rmakara: There is `getCustomerAddressEndpoint()` in `endpoints.ts`, that we may use. :)

packages/shopware-6-client/src/services/customerService.ts
```diff
@@ -0,0 +1,168 @@
...
+  const contextToken = resp.data["sw-context-token"];
+  update({ contextToken });
+  return { contextToken };
+}
+
+/**
+ * End up the session
+ */
+export async function logout(): Promise<void> {
+  await apiService.post(`${getCustomerEndpoint()}/logout`);
```
rmakara: Possible to use `getCustomerLogoutEndpoint()`. :)

packages/shopware-6-client/__tests__/services/CustomerService/getCustomerAddresses.spec.ts
mkucmus: you are right, I replaced it with the existing method call

packages/shopware-6-client/__tests__/services/CustomerService/getCustomer.spec.ts
mkucmus: You are totally right, I just left the 400 and promis rejection without any extra info to just break the logic from getCustomer() and end up with unhandled rejection. We will prepare the consistent error throwing/handling in the next few iterations once we discover all the cases :+1:

packages/shopware-6-client/src/services/customerService.ts
mkucmus: sure, fixed

packages/shopware-6-client/__tests__/services/CustomerService/getCustomerAddress.spec.ts
mkucmus: removed

mkucmus:
I'm proud of the hard work we put into this PR. Thanks for your support guys!

PR #78, opened 2019-10-15
Feature/axios api service

patzick:
+ mocks in tests to be independent from api

### DIFF
--- a/packages/shopware-6-client/src/categoryService.ts
+++ b/packages/shopware-6-client/src/categoryService.ts
@@ -1,6 +1,6 @@
-import axios from "axios";
 import { config } from "./settings";
 import { CATEGORY_ENDPOINT } from "./endpoints";
+import { apiService } from "./apiService";
 
 export interface Category {
   name: String;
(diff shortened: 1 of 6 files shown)


packages/shopware-6-client/src/categoryService.ts
```diff
@@ -22,7 +22,7 @@ export interface CategoryService {
 }
 
 const getCategories = async function(): Promise<SearchResult<Category[]>> {
-  const resp = await axios.get(config.endpoint + getCategoryEndpoint());
+  const resp = await apiService.get(config.endpoint + getCategoryEndpoint());
```
rmakara: Maybe we can use `getXEndpoint()` methods to generate full URLs? This way we will not have to use `config.endpoint` in every method. Not required to do that within' that one pull request but it is an idea for another issue. If you agree with that then we can create separate issue for that.

patzick: even better, I added `reloadConfiguration` method, so we don't have to pass endpoint, but we still can :) I just forgot to remove it here, corrected now!

Issue #56, opened 2019-10-10
Create Category Service Interface


mkucmus:
Category service should be using a Connection layer (based on axios library).
Each method of the service will contain specific types of parameters and return typed data (corresponding to data models' interface - #42 )

_It's assumed that the context token is provided on request layer._

### The proposed list of methods:
1. `searchIds(params): <SearchResult<string|uuid[]>` || `findIds(params): <SearchResult<string|uuid[]>` - search only for category ids for specific query
2. `find(params): <Category>` || `findOne(params): <Category>` - fetch the chosen category details
3. `findAll(params):<SearchResult<Category[]>` || `list(params):<SearchResult<Category[]>`- get the list of all categories fit all parameters passed

All the parameters should be converted into query string (or payload) just before sending a request. 
Params converter was described here: #58 

_SearchCriteria interface_ described in #55 

### Parameters:
 Ad. 1. 
```
interface searchIdsParams extends SearchCriteria {
}
```
Ad. 2.
```
interface findParams {
  id: string # or uuid interface if any
}
```

Ad. 3. 
```
interface findAllParams extends SearchCriteria {
}
```

### The list of fields that can be used in the filter or sorting queries within the category:
```
id: string | uuid
afterCategoryId: string | uuid
parentVersionId: string | uuid
mediaId: string | uuid
displayNestedProducts: number # 1 if true, 0 otherwise
autoIncrement: number
level: number
path: string
type: string # page, link, folder
visible: number # 1 if true, 0 otherwise
active: number # 1 if true, 0 otherwise
name: string
cmsPageId: string | uuid
```

patzick:
1. What's the use case for this method?

2. and 3. i'd be for some kind of naming conventions like
- `getCategory, getCategories`
- `fetchCategory, fetchCategories` (my personal favourite, IMO most accurate)
- `findCategory, findCategories`
- `loadCategory, loadCategories`

2 and 3 should also have the same params. Method which returns single object should just add `limit=1` to filters and cast result Array to an Object.

rmakara:
Using entity names in methods like fetchCategory, fetchCategories seems to be easier solution for using it instead of category.findId, category.findIds.

We need to keep in mind that for Category pages there is also cms content to display in each page. There are fields `cmsPageId` and `cmsPage` in `CategoryInterface`. At some point we should think about handling that with single getCategory method or two getCategory and getCmsPage methods.

[...]

PR #80, opened 2019-10-15
Feature/56 create category service interface

rmakara:
**Changes:**
* New `services/categoryService`.
* getCategories method, that may optionally get sort, limit, page params.
* getCategory(categoryId)
* getCategoryWithAssociation(categoryId, associationName) method, that we may use for fetching the categories with products to out eCommerce Test Site. TODO: Come up with consistent solution for adding nested objects in responses.
* Tests based on Shopware Test Environment API. Not mocked.

**Issue:** 
#56


packages/shopware-6-client/src/services/categoryService.ts
```diff
@@ -1,16 +1,60 @@
...
-export interface SearchResult<T> {
-  total: Number;
-  data: T[];
-}
+const getCategories = async function(
+  pagination?: any,
+  sort?: any,
+  filter?: any
+): Promise<SearchResult<Category[]>> {
+  const resp = await apiService.get(config.endpoint + getCategoryEndpoint(), {
```
rmakara: Maybe we may use `apiService.defaults.baseURL` instead of `config.endpoint` just like you did in tests? `Config.endpoint` is being used during `resetConfiguration` and it loads baseURL in apiService component.

mkucmus: it's deleted now :heavy_check_mark:

Issue #72, opened 2019-10-14
Create Context service interface


mkucmus:
Context service should be using a Connection layer (based on axios library).
Each method of the service will contain specific types of parameters and return typed data, described in this issue on the bottom.

_It's assumed that the current context's token is provided on request layer._

### The list of proposed methods:
```
1. updateContext(updateContextParam): UpdateContextResponse
```

### Available parameters:
```
interface updateContextParam {
  currencyId?: string | uuid
  languageId? : string | uuid
  customerId? : string | uuid
  customerGroupId? : string | uuid
  billingAddressId? : string | uuid
  shippingAddressId? : string | uuid
  paymentMethodId? : string | uuid
  shippingMethodId? : string | uuid
  countryId? : string | uuid
  countryStateId? : string | uuid
}
```

### Response interface
```
interface UpdateContextResponse { 
 sw-context-token: string
}
```

### Method to endpoint map
```
1 updateContext() - PATCH /sales-channel-api/v{version}/context"
```

### Sample request
```
request('PATCH', '/sales-channel-api/v1/context', {
            paymentMethodId: '1ecb5c06-e636-4796-9a5c-3de55cabe137',
        })
```

PR #84, opened 2019-10-15
Feature/72 create context service interface

rmakara:
**Changes:**
* New `services/contextService`.
* `getAvailableCurrencies()` method.
* `setCurrentCurrency()` method.
* `getAvailableLanguages()` method.
* `setCurrentLanguage()` method.
* `getAvailableCountries()` method.
* `getAvailablePaymentMethods()` method.
* `setCurrentShippingMethod()` method.
* `getAvailableShippingMethods()` method.
* `setCurrentShippingMethod()` method.
* Tests.

Patryk also extracted contextToken to global config.

**Issue:** 
#72


packages/shopware-6-client/src/services/contextService.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  return resp.data;
+};
+
+interface ContextService {
+  getCurrencies: () => Promise<SearchResult<Currency[]>>;
+  getLanguages: () => Promise<SearchResult<Language[]>>;
+  getCountries: () => Promise<SearchResult<Country[]>>;
+  getPaymentMethods: () => Promise<SearchResult<PaymentMethod[]>>;
+  getShippingMethods: () => Promise<SearchResult<ShippingMethod[]>>;
```
mkucmus:
```suggestion
  getAvailableCurrencies: () => Promise<SearchResult<Currency[]>>;
  getAvailableLanguages: () => Promise<SearchResult<Language[]>>;
  getAvailableCountries: () => Promise<SearchResult<Country[]>>;
  getPaymentMethods: () => Promise<SearchResult<PaymentMethod[]>>;
  getAvailableShippingMethods: () => Promise<SearchResult<ShippingMethod[]>>;
```
having those methods in cotext is totally fine for me, it's directly related, but under the word `context` we should rather have something related to the customer's current session. 
We could have all those methods with the keyword "available" and also with the "current" word when the SW Core Team makes the methods that responses the current context's details so we don't have any at the moment.

packages/shopware-6-client/src/services/contextService.ts
```diff
@@ -0,0 +1,75 @@
...
+} from "../endpoints";
+import { Country } from "../interfaces/models/system/country/Country";
+import { ShippingMethod } from "../interfaces/models/checkout/shipping/ShippingMethod";
+import { PaymentMethod } from "../interfaces/models/checkout/payment/PaymentMethod";
+import { Language } from "../interfaces/models/framework/language/Language";
+import { SearchResult } from "../interfaces/response/SearchResult";
+
+const getCurrencies = async function(): Promise<SearchResult<Currency[]>> {
+  const resp = await apiService.get(
+    config.endpoint + getContextCurrencyEndpoint()
```
mkucmus: please use template strings like this
```suggestion
    ${config.endpoint}${getContextCurrencyEndpoint()}
```

rmakara: Done :)

rmakara:
**Changes:**
* Adjusted the code and tests just like in Categories and Products.
* I am not sure about splitting the tests of `Set` methods into 2 described. The first one with contextToken given and the second without contextToken given - as we use these setters as logged and not-logged user.

PR #85, opened 2019-10-17
Feature/search criteria

patzick:


packages/shopware-6-client/__tests__/helpers/convertSearchCriteria.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  RangeFilter,
+  MultiFilter
+} from "../../src/interfaces/search/SearchFilter";
+
+describe("SearchConverter - convertSearchCriteria", () => {
+  it("should returns empty object if no params provided", () => {
+    const result = convertSearchCriteria();
+    expect(result).toEqual({});
+  });
+  describe("pagination", () => {
```
rmakara: As only following (below) limits are available in Shopware maybe let's add a limitation for that? Probably we shouldn't create that strict limitation in our code but only catch `400: FRAMEWORK__LIMIT_NOT_ALLOWED` error from API?

Note from Shopware API:
"Please be aware that not all limits are supported. The following limits are currently allowed: 1, 5, 10, 25, 50, 75, 100, 500"

patzick: youre right!
i've added limitations, so you could call `limit: PaginationLimit.ONE` or `limit: 5`
but if you have for eg `limit: 11`, then we don't add it

For future consideration is some method for validation, not sure if we should throw an exception here that parameters are wrong. WDYT @rmakara , @mkucmus ?

rmakara: Writing a comment for historical purposes, good insight from @mkucmus: "As Shopware 6 is an open source, that means everyone can pull it and edit the default / official limits".

Issue #73, opened 2019-10-14
Create checkout service interface


mkucmus:
Checkout service should be using a Connection layer (based on axios library).
Each method of the service will contain specific types of parameters and return typed data, described in this issue on the bottom.

_Custom cart Interfaces listed at the bottom of the issue._

_It's assumed that the current context`s token is provided on request layer._

_consistent error response should be described in separated issue and used globally_

### The list of proposed methods:
```
1. createOrder(ExtraTokenParam?): Order # an Order interface is taken from #54 
2. createGuestOrder(CreateGuestOrderParam): Order # taken from #54 
3. getDeepLinkOrder(GetDeeplinkParam): Order # taken from #54 
4. payOrder(PayOrderParam): PayOrderResponse | HTTP_NO_CONTENT # described below
```

**All the parameters should be converted into query string (or payload) just before sending a request.**
Checkout Params converter was described here: #TODO

### Available parameters:
```
interface ExtraTokenParam {
  token?: string
  name? : string
}
```

```
interface CreateGuestOrderParam { # some parts of Customer data model's interface. not sure if every key can be used
  email: string
  salutationId: string | uuid
  firstName: string,
  lastName: string,
  billingAddress: BillingAddress
  shippingAddress?: ShippingAddress # shippingAddress interface is taken from #54 
}

```

```
interface GetDeeplinkParam {
  orderId: string | uuid
  accessCode: string
}
```

```
interface PayOrderParam {
  orderId: string | uuid
  finishUrl: string
}
```

### The list of Checkout custom interfaces:
```
interface PayOrderResponse {
 paymentUrl: string
}
```

### Method to endpoint map
```
1. createOrder() - POST /sales-channel-api/v{version}/checkout/order
2. createGuestOrder) - POST /sales-channel-api/v{version}/checkout/guest-order"
3. getDeepLinkOrder() - GET /sales-channel-api/v{version}/checkout/guest-order/{id}
4. payOrder()  - POST /sales-channel-api/v{version}/checkout/order/{orderId}/pay
```

mkucmus:
points 1 and 2 are solved within #357

rmakara:
Closing the issue. `createOrder()` and `createGuestOrder` have been implemented.

Guest orders and payments will be implemented in separate tasks.

Associated:
* #375 
* https://github.com/DivanteLtd/shopware-pwa/issues/481
* https://github.com/DivanteLtd/shopware-pwa/issues/482

Issue #64, opened 2019-10-10
Create Cart service interface


mkucmus:
Cart service should be using a Connection layer (based on axios library).
Each method of the service will contain specific types of parameters and return typed data, described in this issue on the bottom.

_Custom cart Interfaces listed at the bottom of the issue._

_It's assumed that the context token is provided on request layer._

### The list of proposed methods:
```
1 getCart(ExtraTokenParam?):CartResponse<Cart> # Cart interface taken from Cart.ts; creates a new one if doesn't exist
2 createCart(ExtraTokenParam?): CartResponse<Cart>
3 addProduct(addProductToCartParam, ExtraTokenParam?):CartResponse<Cart>
4 addCode(PromotionCodeParam, ExtraTokenParam?): CartResponse<Cart>
5 addLineItem(AddLineItemParam): CartResponse<Cart>
6 removeLineItem(RemoveLineItemParam): CartResponse<Cart>
7 updateLineItem(UpdateLineItemParam,ExtraTokenParam?): CartResponse<Cart>
```

**All the parameters should be converted into query string (or payload) just before sending a request.**
Customer Params converter was described here: #TODO

### Available parameters:
```
interface ExtraTokenParam {
  token?: string
  name? : string
}
```

```
interface addProductToCartParam {
  productId: string | uuid
  quantity: number
}
```

```
interface PromotionCodeParam {
  code: string
}
```

```
interface AddLineItemParam {
  itemId: string | uuid
  quantity?: number
}
```

```
interface RemoveLineItemParam {
  itemId: string | uuid
}
```

```
interface UpdateLineItemParam {
  itemId: string | uuid
  quantity: number
}
```

### The list of Cart's custom interfaces:

interface CartResponse<T> { # decide if using the global one is needed
 data: <T>
}

### Method to endpoint map
```
1 getCart() - GET /sales-channel-api/v{version}/checkout/cart
2 createCart() - POST /sales-channel-api/v{version}/checkout/cart 
3 addProduct() - POST /sales-channel-api/v{version}/checkout/cart/product/{id}
4 addCode() - POST /sales-channel-api/v{version}/checkout/cart/code/{code}
5 addLineItem() - POST /sales-channel-api/v{version}/checkout/cart/line-item/{id}
6 removeLineItem() - DELETE /sales-channel-api/v{version}/checkout/cart/line-item/{id}
7 updateLineItem() - PATCH /sales-channel-api/v{version}/checkout/cart/line-item/{id}
[...]

PR #101, opened 2019-10-22
Feature/64 create cart service

rmakara:
**Changes:** 
* Add CartService.
* Add tests for CartService.
* Fix endpoint for adding products in `endpoints.ts`.
* Add endpoint for adding promotion codes to `endpoints.ts`.

**Additional information:**
That PR does not include checkout methods, that will be the part of #73.

**Issue:**
#64


packages/shopware-6-client/src/services/cartService.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export async function getCart(): Promise<Cart> {
+  const resp = await apiService.get(getCheckoutCartEndpoint());
+
+  return resp.data;
+}
+
+export async function addProduct(
+  productId: string,
+  newQuantity: number
```
patzick: why is it called `newQuantity`? I believe it's just quantity :)
```suggestion
  quantity: number
```

packages/shopware-6-client/src/services/cartService.ts
```diff
@@ -0,0 +1,74 @@
...
+    getCheckoutCartProductEndpoint(productId),
+    params
+  );
+
+  return resp.data;
+}
+
+export async function addLineItem(
+  itemId: string,
+  addQuantity: number
```
patzick:
```suggestion
  quantity: number
```

packages/shopware-6-client/__tests__/services/CartService/updateLineItem.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  it("should call valid endpoint and return cart with no items", async () => {
+    mockedAxios.patch.mockResolvedValueOnce({
+      data: {
+        name: "3a64e872ca404522a2c5d43ebc751e6b",
+        token: "NWDdcRTTWoPk4Ngv13z5NDMMsDFRb9W6",
+        lineItems: [
+          {
+            id: "geawq90a5dab4206843d0vc3sa8wefdf",
+            label: "Red T-Shirt",
```
mkucmus: you can go ahead with faker in a place like this
```
import { commerce } from "faker"
```

```suggestion
            label: commerce.productName(),
```

all the id could be generated by faker as well

packages/shopware-6-client/src/services/cartService.ts
```diff
@@ -0,0 +1,74 @@
...
+  return resp.data;
+}
+
+export async function removeLineItem(itemId: string): Promise<Cart> {
+  const resp = await apiService.delete(getCheckoutCartLineItemEndpoint(itemId));
+
+  return resp.data;
+}
+
+export async function addCode(promotionCode: string): Promise<Cart> {
```
patzick: just for method to be more searchable
```suggestion
export async function addPromotionCode(promotionCode: string): Promise<Cart> {
```

patzick: also - don't we need option `removePromotionCode`? It's pretty common to have it.
Especially when we have multiple codes, but even with one. (Can we have multiple codes?)

packages/shopware-6-client/src/services/cartService.ts
```diff
@@ -0,0 +1,74 @@
...
+  );
+
+  return resp.data;
+}
+
+export async function addLineItem(
+  itemId: string,
+  addQuantity: number
+): Promise<Cart> {
+  let params = { type: "product", quantity: addQuantity };
```
mkucmus: It would be really good to have "product" type taken from enum.

packages/shopware-6-client/__tests__/services/CartService/updateLineItem.spec.ts
patzick: niiiiice! ;D

packages/shopware-6-client/src/services/cartService.ts
```diff
@@ -0,0 +1,74 @@
...
+  getCheckoutCartProductEndpoint,
+  getCheckoutPromotionCodeEndpoint,
+  getCheckoutCartLineItemEndpoint
+} from "../endpoints";
+import { apiService } from "../apiService";
+import { ContextTokenResponse } from "../interfaces/response/ContextTokenResponse";
+
+export async function createCart(): Promise<ContextTokenResponse> {
+  const resp = await apiService.post(getCheckoutCartEndpoint());
+  let contextToken = resp.data["sw-context-token"];
```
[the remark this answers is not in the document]
rmakara: Yep, it is possible when no contextToken is given. Added. Thanks for pointing that. x]

package-lock.json
```diff
@@ -0,0 +1,10486 @@
+{
```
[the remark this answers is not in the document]
rmakara: Deleted, sry ^^

packages/shopware-6-client/src/services/cartService.ts
rmakara: LineItemType enum added. :)

packages/shopware-6-client/src/services/cartService.ts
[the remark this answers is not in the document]
rmakara: Done.

Another topic is about updating contextToken in general. In fact, we are getting contextToken with the 1st API call. No matter which. Even /category and /product endpoints return contextToken in the headers, that we may use.

The only one API call, that would change the contextToken is login().

Maybe we may add a global listener in Axios layer for reading the contextToken from the headers?

packages/shopware-6-client/src/services/cartService.ts
rmakara: Name of the method changes.

For removePromotionCode we may use removeCartLineItem. I will write a JSDoc for that.

rmakara:
@mkucmus @patzick 
Fixes applied. :)

PR #102, opened 2019-10-22
Feature/client use on category page

patzick:
1. availability to use `yarn dev:client` to develop locally and refresh page
2. added defaultPaginationLimit to config and using it when no limit is provided
3. changed getProducts to post
4. added associations in searchQuery
5. show product images on category page

Before merge: update readme.md to adjust for local coding

### DIFF
--- a/test-page/src/views/Category.vue
+++ b/test-page/src/views/Category.vue
@@ -118,7 +118,7 @@
             v-for="(product, i) in products"
             :key="i"
             :title="product.name"
-            image="/img/product_thumb.png"
+            :image="product.media && product.media.length ? product.media[0].media.url : '/img/product_thumb.png'"
             :regular-price="product.calculatedPrice.unitPrice"
             :isOnWishlist="false"
             @click:wishlist="toggleWishlist(i)"
@@ -356,7 +356,9 @@ export default {
     };
   },
   async mounted() {
-    this.productsResponse = await getProducts();
+    this.productsResponse = await getProducts({
+      configuration: { associations: [{ name: "media" }] }
+    });
   },
   computed: {
     products() {
--- a/test-page/src/views/Category.vue
+++ b/test-page/src/views/Category.vue
@@ -118,7 +118,11 @@
             v-for="(product, i) in products"
             :key="i"
             :title="product.name"
-            :image="product.media && product.media.length ? product.media[0].media.url : '/img/product_thumb.png'"
+            :image="
+              product.media && product.media.length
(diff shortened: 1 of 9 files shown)


test-page/src/views/Category.vue
```diff
@@ -117,8 +117,12 @@
           <SfProductCard
             v-for="(product, i) in products"
             :key="i"
-            :title="product.name"
-            image="/img/product_thumb.png"
+            :title="product.name || ''"
```
rmakara: Not a fix request but quick insight.

Getting null names (or any kind of basic data, that is nullable) from Shopware might be surprising for the new storefront developers. Maybe there is some way to extract that logic from the template to a method? When we were working on interfaces for Shopware models we saw so many nullable fields. I have a feeling we will need to add `|| ''` in many places, so if that would be possible we should cover these null appearances in some eye-catching layer to make it easy to find later.

test-page/src/views/Category.vue
```diff
@@ -117,8 +117,12 @@
...
+            :image="
+              product.media && product.media.length
+                ? product.media[0].media.url
+                : '/img/product_thumb.png'
```
rmakara: Not required right now but maybe we may extract getting a default thumbnail to some method? We will use that in product and cms pages also. :)

Edit:
After a quick look, I am not sure if there is a default thumbnail for products with no cover in Shopware. I guess Shopware paints default one with polygons. I did not find that in admin panel. In the SW6 codebase I found only lines for throwing an exception in case of not image found.

image

patzick: yeah, this is only temporary, we need wrapper over this component to take Shopwate product and put it in StorefrontUI item :)
and agree - helpers in this area will be necessary!

test-page/src/views/Category.vue
patzick: It's more like fix we need to add into StorefrontUI, it showed an error, when got null value here :) i just puted it here to avoid errors on dev console

develth:
Is it possible to configure the pagination Limit in I  `shopware-pwa.config`?
Thanks!

patzick:
hey @develth  - it is possible via `defaultPaginationLimit` setting in `shopware-pwa.config.js`

develth:
HI @patzick  - i tried - also a build afterwards, but didn´t changed the amount of items

develth:
Ah - the Products are from CmsBlock productListing. So its overridden by data. my fault. thanks!

PR #112, opened 2019-10-28
feat(error): add error response interface

mkucmus:
It adds a common interface of response when the exception is thrown on API side.

### DIFF
--- a/packages/shopware-6-client/src/interfaces/response/Error.ts
+++ b/packages/shopware-6-client/src/interfaces/response/Error.ts
@@ -10,7 +10,7 @@ interface Error {
     function: string;
     class: string;
     type: string;
-    args: Array<any>;
+    args: any[];
   }>;
 }
 
--- a/packages/shopware-6-client/src/interfaces/response/Error.ts
+++ b/packages/shopware-6-client/src/interfaces/response/Error.ts
@@ -15,5 +15,5 @@ interface Error {
 }
 
 export interface ErrorResponse {
-  errors: Array<Error>;
+  errors: Error[];
 }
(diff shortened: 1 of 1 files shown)


packages/shopware-6-client/src/interfaces/response/Error.ts
```diff
@@ -0,0 +1,19 @@
...
+    line: number;
+    function: string;
+    class: string;
+    type: string;
+    args: Array<any>;
+  }>;
+}
+
+export interface ErrorResponse {
+  errors: Array<Error>;
```
patzick:
```suggestion
  errors: Error[];
```

packages/shopware-6-client/src/interfaces/response/Error.ts
```diff
@@ -0,0 +1,19 @@
+interface Error {
+  status: string;
+  code: string;
+  title: string;
+  detail: string;
+  meta: any;
+  trace?: Array<{
```
patzick: could you move it to separate interface?
```suggestion
  trace?: ErrorTrace[];
```

packages/shopware-6-client/src/interfaces/response/Error.ts
```diff
@@ -0,0 +1,19 @@
...
+    file: string;
+    line: number;
+    function: string;
+    class: string;
+    type: string;
+    args: Array<any>;
```
patzick: isn't it a string array?
```suggestion
    args: any[];
```

mkucmus:
Fixes were added. All checks are green :+1:

PR #125, opened 2019-10-30
feat(product-page): hydrate template with real data

mkucmus:


test-page/src/views/Category.vue
```diff
@@ -125,7 +125,7 @@
             "
             :regular-price="product.calculatedPrice.unitPrice"
             :isOnWishlist="false"
-            @click:wishlist="toggleWishlist(i)"
+            @click:wishlist="$router.push({name: 'product', params: {id: product.id}})"
```
mkucmus: temporary workaround, just ignore it

patzick: we don't want to change route on withlist button click :P

test-page/src/views/Product.vue
```diff
@@ -451,9 +390,145 @@ export default {
...
+    },
+    sizes() {
+      if (!this.product && !this.product.children) {
+        return []
+      }
+
+      return this.extractOptions("size")
+    },
+    reviews() {
+      return this.product.productReviews
```
patzick: you're checking `reviews.length` much, so better ensure it always is at least an empty array
```suggestion
      return this.product.productReviews || []
```

test-page/src/views/Category.vue
mkucmus: It's fine, I restored it

Issue #129, opened 2019-11-05
Add first CMS components to Shopware-PWA project


rmakara:
Already implemented:
- [x] vimeo-video
- [x] youtube-video
- [x] text-teaser
- [x] text-two-column
- [x] text-three-column
- [x] - product-listing
- [x] - category-navigation
- [x] - text-hero

TODO

- [x] image-text
- [x] image-text-row 
- [x] image
- [x] product-image slots

rmakara:
Probably most done by @mkucmus 

To verify by @mkucmus

PR #138, opened 2019-11-07
cms page components

mkucmus:
#129


patzick:
@mkucmus we will continue CR and feature in #126

PR #148, opened 2019-11-13
feat(rfc): describe filters and sorting response format.

mkucmus:
Co-authored-by: rmakara <rafal.makara@gmail.com>


rfc/Resolver/CategoryForHeadless.ts
```diff
@@ -1,62 +1,134 @@
...
+  availableSorting: Sort // sample below within avaiableSortingResponseExample
+  availableFiltering: Array<OptionFilter | RangeFilter | BooleanFilter> // sample below within avaiableFiltersResponseExample
+  cmsPage: CmsPage | null
+}
 
-  cmsPage: CmsPage | null;
+// <SORTING>
+interface Sort {
+  code: string
+  value: string // field name to sort by
```
patzick: Isn't it same as `code`? Can they be different?

rfc/Resolver/CategoryForHeadless.ts
```diff
@@ -1,62 +1,134 @@
...
+  label: string // translated label
+  default: string // desc
```
patzick:
```suggestion
  default: string
```

rfc/Resolver/CategoryForHeadless.ts
mkucmus: value could also contain "-" minus sign. can be prefixed with it. but now I don't see any use case when we could use it instead of the value...

PR #149, opened 2019-11-13
feat(category): move category into nuxt structures WIP

mkucmus:


packages/default-theme/layouts/default.vue
```diff
@@ -1,33 +1,20 @@
 <template>
   <div>
-    <div id="nav">
```
patzick: please revert this file, for now, we will have here navigation and probably breadcrumbs

packages/shopware-6-client/src/index.ts
```diff
@@ -1,12 +1,14 @@
 import { ClientSettings, setupConfig, updateConfig } from "./settings";
 import { reloadConfiguration } from "./apiService";
 
-export { config } from "./settings";
+export { config } from "./settings";import { Cart } from "../interfaces/models/checkout/cart/Cart";
```
patzick: I believe this import is unnecessary here
```suggestion
export { config } from "./settings";
```

packages/shopware-6-client/src/services/navigationService.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  getNavigationEndpoint
+} from "../endpoints";
+import { apiService } from "../apiService";
+
+interface NavigationResponse {
```
patzick:
```suggestion
export interface NavigationResponse {
```

packages/shopware-6-client/src/services/navigationService.ts
```diff
@@ -0,0 +1,31 @@
...
+  count: number,
+  elements: NaviationElement[]
+}
+
+interface NaviationElement {
```
patzick:
```suggestion
export interface NavigationElement {
```

packages/shopware-6-client/src/services/navigationService.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  getNavigationEndpoint
+} from "../endpoints";
+import { apiService } from "../apiService";
+
+interface NavigationResponse {
+  count: number,
+  elements: NaviationElement[]
```
patzick:
```suggestion
  elements: NavigationElement[]
```

packages/shopware-6-client/src/services/navigationService.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+interface NaviationElement {
+  id: string
+  path: string
+  name: string
+  children: NaviationElement[] | null,
```
patzick:
```suggestion
  children: NavigationElement[] | null,
```

packages/shopware-6-client/src/services/navigationService.ts
```diff
@@ -0,0 +1,31 @@
...
+  count: number,
+  level: number,
+  extensions: any[]
+}
+
+
+export async function getNavigation(
```
patzick: always try to take params as an object, then it's easy to extend and change the interface without breaking changes. Also, order of params doesn't matter then.

packages/default-theme/layouts/default.vue
mkucmus: moved to another component and used here

PR #164, opened 2019-11-25
Feat/nuxt product

mkucmus:


packages/composables/src/hooks/useProduct/SwHelpers/getMainImageUrl.ts
```diff
@@ -0,0 +1,3 @@
+export default function getMainImageUrl(product, defaultUrl= "/img/product_thumb.png"){
```
patzick:
```suggestion
export default function getMainImageUrl(product){
```
it would be really painful to always provide fallback here, maybe later we would add this to config and then use here :)

packages/composables/src/hooks/useProduct/SwHelpers/getMainImageUrl.ts
```diff
@@ -0,0 +1,3 @@
...
+  return (product.cover && product.cover.media) ? product.cover.media.url : defaultUrl
```
patzick: what if product is null? :)

packages/helpers/__tests__/product/isProductSimple.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { isProductSimple } from "@shopware-pwa/helpers";
```
patzick: all helpers tests should cover more cases like:
- `isProductSimple()`
- `isProductSimple({product: null})`
- `isProductSimple({})`

packages/helpers/__tests__/product/getProductRegularPrice.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { getProductRegularPrice } from "@shopware-pwa/helpers";
+
+describe("Shopware helpers - getProductRegularPrice", () => {
+
+  it("should return first gross price from nested price object", () => {
+    const productWithPrice: any = {
+      price: [
+        {
```
patzick: what if there are more price objects?

packages/helpers/src/product/getProductOptions.ts
```diff
@@ -0,0 +1,33 @@
+import { Product } from "@shopware-pwa/shopware-6-client";
+import { UiProductOption } from "packages/global";
+
+interface Parameters {
+  product: Product
+  attribute: string
+}
+
+export default function getProductOptions(params: Parameters): UiProductOption[]{
```
patzick:
```suggestion
export function getProductVariantsOptions({product, attribute}:{product?:Product, attribute?:string} = {}): UiProductOption[]{
```

packages/helpers/__tests__/product/getProductRegularPrice.spec.ts
mkucmus: discussed during the call

patzick:
Good job, plenty of code! ;D

rmakara:
@mkucmus <3 I simply cannot believe it happened! Wonderful code!

PR #185, opened 2019-12-03
fix(helpers): fix for parentId same as id on parent products

patzick:


packages/helpers/src/product/isProductSimple.ts
```diff
@@ -3,5 +3,7 @@ import { Product } from "@shopware-pwa/shopware-6-client";
 export function isProductSimple({
   product
 }: { product?: Product } = {}): boolean {
-  return product ? !!product.parentId : false;
+  return product
+    ? !!product.parentId && product.id !== product.parentId
```
mkucmus: I hope it's a temporary workaround :)

PR #194, opened 2019-12-06
refactor: create component for product gallery

### LINKED ISSUE #15: I want to review the product gallery So I can see how the product looks like
**Context**
In product page we want to see all the photos of the product. We can hold many photos to attract the customer with the produduct more. Images are stored as MediaEntity in Shopware. We should figure it out if we need to fetch the whole media or basic image infromation only.

**Acceptance criteria**
* I can see a gallery with 1 image in the product page.
* I can see a gallery with many images in the product page.
* I can switch the images.
* I can see image title when I put cursor above the image.
* I can read valid ALT parameter.

krskibin:
Closes #15 
- Separated the product gallery from product view
- Added slots for that component to easily customize the gallery
- Fixed product photo displaying

### DIFF
--- a/packages/default-theme/components/cms/elements/SwProductGallery.vue
+++ b/packages/default-theme/components/cms/elements/SwProductGallery.vue
@@ -1,0 +1,100 @@
+<template>
+  <div class="sw-product-gallery gallery">
+    <template v-if="mediaGallery.length > 0">
+      <slot name="desktop-galery" v-bind="mediaGallery">
+        <div class="gallery__desktop" v-for="image in mediaGallery" :key="image.big.url">
+          <SfImage
+            :src="image.big.url"
+            class="cover-image desktop-only"
+          />
+        </div>
+      </slot>
+    </template>
+    <div class="gallery__mobile mobile-only" v-if="mediaGallery.length > 0">
+      <slot name="mobile-galery" v-bind="mediaGallery">
+        <SfGallery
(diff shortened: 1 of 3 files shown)


packages/default-theme/components/cms/elements/SwProductGallery.vue
```diff
@@ -0,0 +1,100 @@
...
+import { SfImage, SfGallery } from '@storefront-ui/vue'
+import { getProductMainImageUrl, getProductMediaGallery } from '@shopware-pwa/helpers'
+
+export default {
+  components: { SfImage, SfGallery },
+  props: {
+    product: {
+      type: Object,
+      default: () => ({}),
+      required: true
```
patzick: the default value is what protects us when the prop is not provided. We should never use the required properties, unless it's completely necessary.
```suggestion
```

packages/default-theme/components/cms/elements/SwProductGallery.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <div class="sw-product-gallery gallery">
+    <template v-if="mediaGallery.length > 0">
+      <slot name="desktop-galery" v-bind="mediaGallery">
+        <div class="gallery__desktop" v-for="image in mediaGallery" :key="image.big.url">
```
patzick: what if `big` would be null?

PR #195, opened 2019-12-08
fix(theme): update contextToken only if has changed

patzick:


patzick:
Resolved problem with first render after on Server. Nuxt was trying to set a cookie after the request sent which caused a problem with initial render.
At first, I thought it was caused by modifying global Axios in response interceptor, but cookies were the real reason. They can fail silently with no impact on the application.

Issue #25, opened 2019-10-02
I want to logout 


rmakara:
**Context**
In #24 I logged in. Now I want to logout by calling the valid Shopware API endpoint, which terminates the sw-context-token

**Acceptance criteria**
* When I am logged in I can see a Logout button.
* When I am not logged in I cannot see a Logout button.
* When I click on the Logout button then I am logged out.
* I can see information that I got logged out.
* All the user-related data is being refreshed from Shopware API.

PR #201, opened 2019-12-10
Feat/user management

### LINKED ISSUE #24: I want to login as a customer So I can access My Account details and continue shopping as recognized user
**Context**
Users in Shopware are based on `sw-context-token`. After successful login user we should get the new sw-context-token, which will be added to all the next requests.

**Acceptance Criteria**
* I can visit Login page.
* I can enter my username and password.
* I can see login button in the Login Page.
* When I click login having empty fields then I can see validation error.
* When I entered the correct data I am being logged in.
* New sw-context-token is assigned.
* When I entered invalid data then I can see an error message.
* Icon / text in a header for Login is changed to Logout context.

patzick:
closes #24 #25

### DIFF
--- a/packages/shopware-6-client/__tests__/services/CustomerService/getCustomer.spec.ts
+++ b/packages/shopware-6-client/__tests__/services/CustomerService/getCustomer.spec.ts
@@ -18,19 +18,36 @@ describe("CustomerService - getCustomer", () => {
     });
   });
 
-  it("should an current customer's data - using correct token", async () => {
+  it("should return current customer's data - using correct token", async () => {
     mockedAxios.get.mockResolvedValueOnce({
       data: { data: { id: "c370eb5cd1df4d4dbcc78f055b693e79" } }
+    });
+    const result: any = await getCustomer();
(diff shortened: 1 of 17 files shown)


packages/shopware-6-client/__tests__/services/CustomerService/getCustomer.spec.ts
```diff
@@ -18,19 +18,36 @@ describe("CustomerService - getCustomer", () => {
...
     expect(result.id).toEqual("c370eb5cd1df4d4dbcc78f055b693e79");
   });
 
-  it("should an current customer's data - using incorrect token", async () => {
-    mockedAxios.get.mockRejectedValue(new Error());
-    expect(getCustomer()).rejects.toThrow();
+  it("should return null when user not logged in", async () => {
+    mockedAxios.get.mockRejectedValueOnce({
+      response: {
+        status: 403
```
mkucmus: finally we are using proper http codes, not 500 on everything :+1: you know what I mean :)

Issue #230, opened 2019-12-19
Use Login Form from Storefront UI


rmakara:
**Context**
We added Login and Logout possibility in the previous tasks (#25 #24). Now we want to use the styles from Storefront UI lib so it will look better.

**Dependency**
* https://github.com/DivanteLtd/storefront-ui/pull/508

**Acceptance Criteria**
* Example Login from Storefront UI is used.
* Validation does not appear just after poping up a modal.
* Validation does not appear just after poping up a modal for the second time.
* Validation error message appear when invalid data is given.
* Reset passwords and Register links do not need to work.
* Closing the modal window works.

Issue #33, opened 2019-10-02
I want to reset the password So I can get a new one when I forgot the current one


rmakara:
**Context**
In the previous tasks we made it possible to register user account, login and logout. Now we want to let users to reset theirs password if they forgot it.

**Acceptance Criteria**
* In the login page I can see the possibility to reset the password.
* I can enter my login / email of the account that I want to reset.
* I can successfully click the Reset My Password button.
* According to the Shopware logic I probably get an email from Shopware.
* After I change my password I can successfuly login.

PR #245, opened 2020-01-07
refactor(theme): use storefront ui login form #230

### LINKED ISSUE #26: I want to register as a new customer So I can later login to My Account
**Context**
In #24 and #25 tasks we managed to login and logout to/from existing user account. Now we want to let users to create a new account.

**Acceptance Criteria**
* I can go to Registration Page.
* I can successfully create a new user account.

**Not in the scope**
* I can fill all the necessary infromation for Shopware: Salutation, First name, Last name, Email, Street, Zipcode, City, Country.
* Validation for the fields work well.
* When I do not fill the form correctly and I use Register button then I can see validation errors.

krskibin:
Closes #26, #230, #33


package.json
```diff
@@ -78,5 +78,8 @@
   },
   "engines": {
     "node": "10.x"
+  },
+  "dependencies": {
```
patzick: wrong package, it should go into default-theme package

packages/default-theme/components/SwLogin.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+  <div class="sw-login">
+    <div class="form sw-login__form">
+      <h2 class="sw-login__header">Log in</h2>
+      <SfAlert class="sw-login__alert" v-if="error" type="danger" :message="error" />
+      <SfInput
+        v-model="$v.login.$model"
```
patzick:
```suggestion
        v-model="login"
```

packages/default-theme/components/SwLogin.vue
```diff
@@ -0,0 +1,103 @@
...
+        name="login"
+        label="Your login"
+        class="form__input"
+        :valid="!$v.login.$error"
+        :disabled="isLoading"
+        errorMessage="Login is required"
```
patzick: there should be also `onblur` method which invoked `touch` on this field
and same with other field

packages/default-theme/components/SwLogin.vue
```diff
@@ -0,0 +1,103 @@
...
+      />
+      <SfInput
+        v-model="$v.password.$model"
```
patzick:
```suggestion
        v-model="password"
```

packages/default-theme/components/SwLogin.vue
```diff
@@ -0,0 +1,103 @@
...
+      required,
+    },
+    password: {
+      required,
+    }
+  },
+  methods: {
+    async invokeLogin() {
+      this.$v.$touch()
+      if(this.$v.$anyError) {
```
patzick:
```suggestion
      if(this.$v.$invalid {
```
look at this example for the difference https://vuelidate.js.org/#sub-error-vs-any-error

packages/default-theme/components/SwResetPassword.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <div class="sw-reset-password">
+    <div class="form sw-reset-password__form">
+      <h2 class="sw-reset-password__header">Reset password</h2>
+      <SfAlert class="sw-reset-password__alert" v-if="error" type="danger" :message="error" />
+      <SfInput
+        v-model="$v.email.$model"
```
patzick:
```suggestion
        v-model="email"
```
and + onblur

packages/composables/src/hooks/useUser.ts
```diff
@@ -32,6 +34,24 @@ interface UseUser {
...
+  salutation: string;
+  street: string;
+  city: string;
+  zipcode: string;
+  country: string;
+  salutationId: string;
+  countryId: string;
+  email: string;
+  title?: string;
+  birthdayYear?: number;
```
patzick: why cannot it be just a date?

packages/composables/src/hooks/useUser.ts
```diff
@@ -32,6 +34,24 @@ interface UseUser {
...
 
+interface RegisterCustomerParams {
+  firstName: string;
+  lastName: string;
+  password: string;
+  salutation: string;
+  street: string;
+  city: string;
+  zipcode: string;
+  country: string;
```
patzick: Do we need both country and countryId?

packages/composables/src/hooks/useUser.ts
```diff
@@ -60,6 +80,35 @@ export const useUser = (): UseUser => {
     }
   };
 
+  const register = async (params: RegisterCustomerParams): Promise<boolean> => {
+    loading.value = true;
+    error.value = null;
+    try {
+      await apiRegister({
```
patzick: is RegisterCustomerParams different from apiRegister? couldn't it just be:
```suggestion
      await apiRegister(params);
```
?

packages/default-theme/components/SwLogin.vue
```diff
@@ -0,0 +1,115 @@
...
+    },
+    password: {
+      required
+    }
+  },
+  methods: {
+    async invokeLogin() {
+      this.$v.$touch()
+      if (this.$v.$invalid) {
+        this.error = 'Form is not valid'
```
patzick: not user-friendly message
```suggestion
        this.error = 'Please provide your login and password.'
```

patzick: and it should not be set here, but in alert and visible only when $v.$error
setting this here may cause problems with a correct error message and not cleaning it properly as most of the logic is in composable

packages/default-theme/components/SwRegister.vue
```diff
@@ -0,0 +1,269 @@
...
+    },
+    city: {
+      required
+    }
+  },
+  methods: {
+    async invokeRegister() {
+      this.$v.$touch()
+      if (this.$v.$invalid) {
+        this.error = 'Form is not valid'
```
patzick: same as with login logic

packages/default-theme/components/modals/SwLoginModal.vue
```diff
@@ -1,53 +1,60 @@
...
 </template>
-<script>
 
-import { SfIcon, SfModal, SfInput, SfButton, SfLoader, SfAlert } from '@storefront-ui/vue'
-import { VuelidateMixin } from '@vuelidate/core'
-import { required } from '@vuelidate/validators'
+<script>
+import { SfButton, SfModal, SfAlert } from '@storefront-ui/vue'
 import { useUser } from '@shopware-pwa/composables'
+import SwRegister from '../SwRegister'
```
patzick: Login is default, but other components should be loaded on demand
```suggestion
const SwRegister = () => import("./SwRegister");
```

packages/default-theme/components/modals/SwLoginModal.vue
```diff
@@ -1,53 +1,60 @@
...
+import SwResetPassword from '../SwResetPassword'
```
patzick:
```suggestion
const SwResetPassword = () => import("./SwResetPassword");
```

packages/composables/src/hooks/useUser.ts
krskibin: Maybe I would reduce the whole interface to fit only required fields for registration

packages/composables/src/hooks/useUser.ts
krskibin: There are different, I change some structure of that interface to fill the billing address also
```js
        {
        firstName: params.firstName,
        lastName: params.lastName,
       salutationId: params.salutationId,
       // ...
        billingAddress: {
          firstName: params.firstName,
          salutationId: params.salutationId,
          lastName: params.lastName,
         // ..
        }
```
Do you have any other idea to prevent from code redundancy?

packages/composables/src/hooks/useUser.ts
patzick: yes, for now, add only required ones. I guess countryId is enough. And please keep dates in single fields ;D

packages/default-theme/components/SwLogin.vue
krskibin: Ok, It was temporary solution. Currently after clicking register/login without filling any fields it will show errors individually for each field, so you right that code should be deleted.

packages/composables/src/hooks/useUser.ts
patzick: you should be consistent with API structure, otherwise, when someone would want to add different addresses (billing, shipping) during registration your structure wouldn't allow that.
It all depends on which interface we want to give users for registration. If is different than API clients, then we need to map them.

packages/composables/src/hooks/useUser.ts
krskibin: I've taken it from here
Zrzut ekranu 2020-01-17 o 14 17 15

packages/composables/src/hooks/useUser.ts
krskibin: I assumed that while the registration user provide only minimum informations that are required, and later while the ordering they will fill the rest needed data. In this case I change the interface structure to simplify it and map the rest informations.

krskibin: Ok, I will move that logic to the fronted component

PR #278, opened 2020-01-13
fix: proper attribute value handling (color prop)

mkucmus:
adds "color" property conditionally.

### DIFF
--- a/packages/default-theme/package.json
+++ b/packages/default-theme/package.json
@@ -21,7 +21,6 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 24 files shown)


packages/default-theme/package.json
```diff
@@ -21,6 +21,7 @@
     }
   },
   "dependencies": {
+    "@babel/preset-env": "^7.8.2",
```
patzick: why is this needed?

mkucmus: It was probably caused obsolete node_modules.

PR #280, opened 2020-01-14
Feat/cli tool

patzick:
CLI package. Instruction for later put in README

To test it follow instructions in `PWA setup` section in CHEATSHEET.md file

Progress of this PR:
- [x] generate CLI project
- [x] init Nuxt project
- [x] update nuxt config
- [x] update nuxt plugins
- [x] add packages 
- [x] prepare default files
- [x] add more info about what's currently going on
- [x] add config file with shopware instance info

In scope after this PR
- download and put plugin files
- prepare to publish CLI package
- collect all TODOs from methods and put them into issues


patzick:
First `init` run:
image

Second `init` run (and every other):
image

Running `shopware-pwa dev`:
image

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,23 @@
...
+  name: "generate",
+  alias: ["g"],
+  run: async (toolbox: GluegunToolbox) => {
+    const {
+      parameters,
+      template: { generate },
+      print: { info }
+    } = toolbox;
+
+    const name = parameters.first;
```
mkucmus: it generates undefined-model.ts if no filename provided

patzick: yeah, it's just a showcase command, I will remove it later so I haven't mentioned it in documentation :)

Issue #298, opened 2020-01-17
Fix Category Description


rmakara:
@akrajna  to fill issue description

PR #302, opened 2020-01-17
feat: show proper product names, add product reviews

### LINKED ISSUE #300: Implement workaround for parent products in Category Page

mkucmus:
closes #300 and #298 

to remember: handle same situation on page resolver when is ready

### DIFF
--- a/packages/default-theme/components/SwProductCard.vue
+++ b/packages/default-theme/components/SwProductCard.vue
@@ -1,11 +1,13 @@
 <template>
   <SfProductCard
-    :title="product.name || ''"
+    :title="getName"
     :image="getImageUrl"
     :special-price="getSpecialPrice"
+    :regular-price="getRegularPrice"
+    :max-rating="5"
+    :score-rating="getProductRating"
     :image-width="216"
     :image-height="326"
-    :regular-price="getRegularPrice"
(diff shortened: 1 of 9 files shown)


packages/default-theme/components/SwProductCard.vue
```diff
@@ -59,6 +62,16 @@ export default {
     }
   },
   computed: {
+    getName() {
```
patzick: This definitely should be placed inside product helpers :)

packages/default-theme/components/SwProductCard.vue
```diff
@@ -41,7 +44,7 @@ export default {
   },
   setup({ product }) {
     const { addToCart, quantity, getStock, isInCart } = useAddToCart(product)
-
+    //console.warn('setup', product.id, product.options);
```
patzick: We can remove that kind of comments

packages/default-theme/components/SwProductCard.vue
mkucmus: done

packages/default-theme/components/SwProductCard.vue
mkucmus: :+1: removed

PR #310, opened 2020-01-21
fix(helpers): remove field check and use explicit field for product name

elkmod:


mkucmus:
@elkmod now it's clear we need to use always translated name. 
please adjust the tests and will be fine :+1:

PR #327, opened 2020-01-24
Fix/login process

mkucmus:
- perform redirect after the login process succeeds 
- distinguish the 401 response
- change some client errors handling: page, login


packages/composables/src/hooks/useUser.ts
```diff
@@ -61,7 +63,9 @@ export const useUser = (): UseUser => {
       await apiLogin({ username, password });
       return true;
     } catch (e) {
-      error.value = e.message;
+      error.value =
+        (e.response && e.response.status === 401 && MESSAGE_ON_LOGIN_401) ||
```
mkucmus: @patzick have a look here

patzick: @mkucmus I think `login` method returns only true/false we should stick to that.
True - successful login attempts, false - bad credentials; and global catch on API client - 500, display additional notification or something :) 
invoking here strings like error-responses can provide further problems with translations, as all visible parts should be defined in theme

mkucmus: sure, as we all know how to deal with errors and hot handling them - I've reverted things you mentioned above. 
the global handler is another issue and approach we are heading to.

PR #472, opened 2020-03-10
feat: add basic healthcheck endpoint for CI/CD purposes

mkucmus:
adds an endpoint at /_healthcheck - returns 204 NO CONTENT if the application is up and running


mkucmus:
> If we're adding health check we should have some info about version, build timestamp and maybe running time?
> WDYT folks @mkucmus, @elkmod ?

I agree, we need to consider those things to be contained:
1. shopware-pwa version
2. shopware-api version compatibility
3. build timestamp

elkmod:
Let's discuss on the review tomorrow, @mkucmus @patzick?

patzick:
I'm closing that for now. The feature will be done  in #561

PR #478, opened 2020-03-10
fix: 502 on composition-api error

mkucmus:


packages/default-theme/pages/_.vue
```diff
@@ -49,8 +49,7 @@ export default {
       cmsPageName: name,
       page: unwrappedPage,
       breadcrumbs,
-      cmsPage,
-      category
```
patzick: shouldn't also try to import it from the line:
`const {search, page, category, error} = useCms()`

mkucmus: you're right, fixed!

Issue #507, opened 2020-03-17
[Checkout] I want to create an account from Personal Details step


patzick:
Personal details step creates account if the checkbox is marked

## Acceptance criteria
- [ ] Continue button registers user
- [ ] registration logs user in
- [ ] user goes to shipping step after a successful operation
- [ ] in case of error proper field is marked or banner with a clear message what went wrong is displayed

PR #518, opened 2020-03-19
feat(helpers): extract messages from API error in case of 400

mkucmus:
Part of work of #507 

covers only helpers package

## Checklist

- [x] I followed code and contributing guidelines
- [x] I wrote all needed tests

### DIFF
--- a/packages/helpers/src/error/index.ts
+++ b/packages/helpers/src/error/index.ts
@@ -1,0 +1,22 @@
+import { ShopwareError } from "@shopware-pwa/commons/interfaces/errors/ApiError";
+
+/**
+ * Get the messages from the API response (array of ShopwareErrors)
+ */
+export function getMessagesFromErrorsArray(
+  errors: ShopwareError[]
+): String[] | undefined {
+  if (!errors || !Array.isArray(errors) || !errors.length) {
+    return;
+  }
+  // return a message only if detail and pointer propery is provided in upcoming ShopwareError
+  return errors
+    .map(
+      ({ detail, source }) =>
+        detail &&
+        source &&
+        source.pointer &&
+        `${source.pointer.substring(1)}: ${detail}`
+    )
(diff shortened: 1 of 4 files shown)


packages/helpers/src/error/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Get the messages from the API response (array of ShopwareErrors)
+ */
+export function getMessagesFromErrorsArray(
+  errors: ShopwareError[]
+): String[] | undefined {
+  if (!errors || !Array.isArray(errors) || !errors.length) {
+    return;
+  }
```
patzick: would be nicer to always return an array :)
+ return primitive string
```suggestion
export function getMessagesFromErrorsArray(
  errors: ShopwareError[]
): string[] {
  if (!errors?.length || !Array.isArray(errors)) {
    return [];
  }
```

packages/helpers/src/error/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  // return a message only if detail and pointer propery is provided in upcoming ShopwareError
+  return errors
+    .map(
+      ({ detail, source }) =>
+        detail &&
+        source &&
+        source.pointer &&
+        `${source.pointer.substring(1)}: ${detail}`
+    )
+    .filter(message => message);
```
patzick: this is not doing anything :)
```suggestion
```

mkucmus: the output of .map function is possible of type "undefined" (due to condition) so it's just filtering out all the falsy values :) but now I can see the better solution would be .filter(String) to get only the strings...

packages/helpers/src/error/index.ts
mkucmus: sure! it's more elegant :+1:

patzick:
great work! 👍

Issue #581, opened 2020-04-03
docs force task


patzick:
## Acceptance criteria
- [x] generate as many docs as we can in 90 mins

patzick:
Done with 3 PRs :)

PR #587, opened 2020-04-06
docs: canary installation

patzick:
## Changes

Installation of canary version

image

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ PS: Check [StorefrontUI](https://github.com/DivanteLtd/storefront-ui/) - our UI 
   </tbody>
 </table>
 
+See [the documentation](https://shopware-pwa-docs.netlify.com/#introduction-to-shopware-pwa) to get started.
+
 ## Video demo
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,6 +30,8 @@ PS: Check [StorefrontUI](https://github.com/DivanteLtd/storefront-ui/) - our UI
   </tbody>
 </table>
 
+See [the documentation](https://shopware-pwa-docs.netlify.com/#introduction-to-shopware-pwa) to get started.
```
Michal-Dziedzinski, line 33: Better place :)

README.md
```diff
@@ -52,6 +50,14 @@ See [the documentation](https://shopware-pwa-docs.netlify.com/#introduction-to-s
 npm install -g @shopware-pwa/cli
 ```
 
+::: tip Canary version  
+Currently, we're releasing a `canary` version per every push to `master` branch, so in order to have newest changes and fixes just install it like this:
```
Michal-Dziedzinski, line 54: Maybe we should mention that this version may contain some issues?

patzick: I thought about it, but in some cases, it's more stable than alpha release ;D
When we move to Beta stage we'll add that kind of warning, as safer then just use Beta

patzick:
closes #581

Issue #8, opened 2019-10-02
I want to use Mega Menu for navigation So I can quickly get to nested categories


rmakara:
**Context**
Currently, I can use the navigation in the header (SfHeader component) to browse level 0 categories. Also, I can use the left-hand-side category tree in category pages.

Now, I want to use the Mega Menu component to quickly get access to nested categories.

**Acceptance criteria**
* Mega Menu component from Storefront UI is used.
* I can see level = 0 categories.
* I can see level = 1 categories.
* I can see level = 2 categories - TO REFINE / VERIFY.
* When I click on the link to a specific category then I am redirected to the correct one

**Blocks**
* #427 CMS content in mega menu

rmakara:
@patzick @mkucmus 
Long time no see. Would be cool to add mega menu in the next sprints. :)

CthulhuDev:
https://storybook.storefrontui.io/?path=/story/organisms-megamenu--with-sfheader

MegaMenu already has multiple depths. You need to override the default theme and change the props of the megamenu

PR #593, opened 2020-04-08
feat: add MegaMenu component

Michal-Dziedzinski:
## Changes

SfMegaMenu implemented as new TopNavigation menu

megamenu

### Checklist


packages/default-theme/components/SwTopNavigation.vue
```diff
@@ -89,46 +85,60 @@ import {
...
     const { count } = useCart()
     const { toggleSidebar } = useCartSidebar()
     const { toggleModal } = useUserLoginModal()
     const { search: fulltextSearch } = useProductSearch()
 
+    const navigationElements = reactive([])
+    const hoveredNavigationItem = ref('')
+
+    onMounted(async () => {
+      const { children } = await getNavigation({
```
mkucmus, line 106: Is it possible to move the logic into `useNavigation` composable?

packages/default-theme/components/SwTopNavigation.vue
```diff
@@ -89,46 +85,60 @@ import {
   useProductSearch
 } from '@shopware-pwa/composables'
 import SwLoginModal from '@shopware-pwa/default-theme/components/modals/SwLoginModal'
+import SwMegaMenu from '@shopware-pwa/default-theme/components/SwMegaMenu'
+import { ref, reactive, onMounted } from "@vue/composition-api";
+import { getNavigation } from '@shopware-pwa/shopware-6-client'
```
mkucmus, line 90: using shopware-6-client directly is not very common in the theme and we should rather keep it in other layer like composition.

packages/default-theme/components/SwTopNavigation.vue
```diff
@@ -13,21 +13,17 @@
...
-            <a
-              :style="{
-                display: 'flex',
-                alignItems: 'center',
-                height: '100%'
-              }"
-            >
-              {{ routeLabel }}
-            </a>
+          <nuxt-link class="sf-header__link" :to="`/${category.route.path}`">
```
mkucmus, line 23: It would be better to have the `path` and any other deep properties using helpers package. That's the layer to transform or extract SW specific nested data. it could be named for instance `getNavigationElements`, and on the occasion, you could create the interface for global (next projects) purposes ;) so, two birds killed with the one stone :smiley: 
less logic in theme is always better :blue_heart:

packages/default-theme/components/SwTopNavigation.vue
Michal-Dziedzinski: sure, good idea!

packages/default-theme/components/SwTopNavigation.vue
Michal-Dziedzinski: You are completely right

patzick:
even worse on page refresh

image

mbaran-dvnt:
Zrzut ekranu 2020-04-19 o 22 15 14

Let me know if that's what you need.

Michal-Dziedzinski:
@mbaran-dvnt if it is related to https://github.com/DivanteLtd/storefront-ui/issues/1029 based on this screen I can't say if this solves our problem.

akrajna:
If there is a lot of subcategories they are nor all visible e.g. in category Sport there are 8 active
 subcategories but on the page we can see 6 and half. 
MegaMenuSubcategory
MegaMenuSubcategories2

patzick:
closes #8

PR #594, opened 2020-04-08
feat(composables): add logic for checkout and sessionContext

patzick:
Related to 
#481

* useSessionContext
* useCheckout
* createCheckoutStep factory
* remove placeOrder from useCart


packages/composables/src/factories/createCheckoutStep.ts
```diff
@@ -0,0 +1,107 @@
...
+export function createCheckoutStep({
+  stepNumber,
+  stepFields,
+  stepDataUpdated,
+}: {
+  stepNumber: number;
+  stepFields: CheckoutStepFields;
+  stepDataUpdated: (updatedData: CheckoutStepFields) => void;
+}) {
+  const cookies = cookieUniversal();
```
mkucmus, line 49: in the future, we can take under consideration leaving here only the interfaces, and pass the right instance - here we have good example of it - the cookies interface with get and set.

mkucmus: oh, you did it with VuelidateValidation :+1:

packages/composables/src/logic/useCheckout.ts
```diff
@@ -0,0 +1,104 @@
...
+        e
+      );
+      throw e;
+    } finally {
+      await refreshCart();
+    }
+  };
+  const isGuestOrder = computed(() => !isLoggedIn.value);
+
+  const guestOrderParams = computed({
```
mkucmus, line 92: :+1: I like that `computed` ;)

packages/composables/src/logic/useSessionContext.ts
```diff
@@ -0,0 +1,53 @@
...
+      vuexStore.commit("SET_SESSION_CONTEXT", context);
+    } catch (e) {
+      console.error("[UseSessionContext][refreshSessionContext]", e);
+    }
+  };
+
+  const shippingMethod = computed(
+    () => sessionContext.value?.shippingMethod || null
+  );
+  const setShippingMethod = async (shippingMethodId: string) => {
```
mkucmus, line 43: I noticed that we are using sometimes simple scalar parameters and sometimes we choose an object. 
you can see it above in `getPaymentMethods` definition (https://github.com/DivanteLtd/shopware-pwa/pull/594/files#diff-8df68b0e440a45a35ea5f7051afaaff5R62). I'm confused because we don't have any global rule on it :)

packages/composables/src/factories/createCheckoutStep.ts
patzick: we use this lib anyway for setting api-client in next, that's why I moved it here for now; in future, for sure we can do it with other tools as well :)

PR #598, opened 2020-04-10
feat/currency switch #580

### LINKED ISSUE #580: Currency switch
## Acceptance criteria
- [ ] implement view from SFUI
- [ ] implement changing context for change currency
- [ ] invoke page reload after change

Figma:
https://www.figma.com/file/N0Ct95cSAoODNv7zYS01ng/Storefront-UI-%7C-Design-System?node-id=771%3A306

mkucmus:
## Changes
closes #580 

### Checklist


packages/default-theme/components/SwProductDetails.vue
```diff
@@ -156,12 +154,19 @@ export default {
...
+      return regularPrice > specialPrice ? regularPrice : specialPrice
     },
     getSpecialPrice() {
-      const price = getProductSpecialPrice(this.product)
-      return price && '$' + price
+      // remove that logic once the SW6 API returns right data
+      // related: https://github.com/DivanteLtd/shopware-pwa/issues/263
+      const regularPrice = getProductRegularPrice({ product: this.product })
+      const specialPrice = getProductSpecialPrice(this.product)
+      return regularPrice > specialPrice ? specialPrice : regularPrice
```
Michal-Dziedzinski, line 169: Should it works like that? :D I know that it's strange that special price is higher than regular, but it was set as special 🤔

packages/shopware-6-client/src/services/contextService.ts
```diff
@@ -77,7 +77,7 @@ export async function getAvailableCurrencies(): Promise<
 > {
   const resp = await apiService.get(getContextCurrencyEndpoint());
```
Michal-Dziedzinski, line 78:
```suggestion
  const {data}= await apiService.get(getContextCurrencyEndpoint());
```
Maybe destructuryzation?

packages/default-theme/components/SwProductDetails.vue
mkucmus: see the comment above :P will be deleted as soon as we have it on API side fixed ;)

packages/composables/src/hooks/useCurrency.ts
```diff
@@ -0,0 +1,96 @@
...
+  availableCurrencies: Ref<Readonly<any>>;
+  currentCurrency: Ref<Readonly<any>>;
+  currentCurrencySymbol: Ref<Readonly<any>>;
+}
+
+/**
+ * @alpha
+ */
+export const useCurrency = (): UseCurrency => {
+  const localCurrency = reactive(sharedCurrency);
```
patzick, line 28: we'll add here something like 
```ts
const {currency, setCurrency} = useSessionContext()
```

then currency we can export right away, and add only compute for currencySymbol

packages/default-theme/components/SwBottomNavigation.vue
```diff
@@ -54,6 +59,7 @@ import {
   useUserLoginModal
 } from '@shopware-pwa/composables'
 import { PAGE_ACCOUNT } from '../helpers/pages'
+import SwCurrency from './SwCurrency'
```
patzick, line 62: we need to import them by aliases, so `@shopware-pwa/default-theme.....`

packages/default-theme/components/SwCurrency.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+  <div class="container" v-if="activeCurrency && availableCurrencies && availableCurrencies.length">
+    <SfSelect v-model="activeCurrency" :size="availableCurrencies.length" class="container__select">
```
patzick, line 3: availableCurrencies should be loaded after a click on select, so first, loadAvailableCurrencies, before that we just displaying current currency from the context

packages/default-theme/components/SwCurrency.vue
```diff
@@ -0,0 +1,93 @@
...
+      availableCurrencies,
+      currentCurrency,
+      fetchCurrencies,
+      switchCurrency
+      } = useCurrency()
+
+    const activeCurrency = computed({
+      get: () => currentCurrency.value && currentCurrency.value.id,
+      set: (currencyId) => switchCurrency(currencyId).then(() => {
+        if (window) {
```
patzick, line 35: maybe we could try with `this.$router.go()`
https://stackoverflow.com/questions/41301099/do-we-have-router-reload-in-vue-router

packages/default-theme/components/SwProductCard.vue
```diff
@@ -72,16 +72,20 @@ export default {
       return getProductUrl(this.product)
     },
     getRegularPrice() {
+      // remove that logic once the SW6 API returns right data
```
patzick, line 75: please add `TODO:` in these comments, to be easily tracked; great that issues are linked <3

```suggestion
      // TODO: remove that logic once the SW6 API returns right data
```

packages/default-theme/nuxt.config.js
```diff
@@ -30,7 +30,7 @@ export default {
   /*
    ** Plugins to load before mounting the App
    */
-  plugins: ['~/plugins/api-client'],
+  plugins: ['~/plugins/api-client', '~/plugins/price-filter'],
```
patzick, line 33: this has no effect, we can remove it from here
```suggestion
  plugins: [],
```

packages/default-theme/components/SwCurrency.vue
mkucmus: this.$router.go needs to be this aware, but it's not inside nested callbacks

packages/default-theme/components/SwCurrency.vue
mkucmus: SfSelect needs to have all options when the activeCurrency is set. Because there is a reference to one from available currencies stick to the current one ;(

packages/composables/src/hooks/useCurrency.ts
mkucmus: :+1:

packages/default-theme/components/SwBottomNavigation.vue
mkucmus: :+1:

packages/default-theme/components/SwProductCard.vue
mkucmus: :+1:

packages/default-theme/nuxt.config.js
mkucmus: :+1:

PR #630, opened 2020-04-20
refactor(cms): cms reorganisation

### LINKED ISSUE #234: Prepare CMS Pages package

patzick:
## Changes

closes #234 

Example of section from theme, overrides block from plugin and overrides element from final project 

image

And plus Warning if CMS config is not correct - no destination file
image


packages/default-theme/components/cms/slots/CmsSlotVideoVimeo.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="sw-vimeo-video">
+  <div class="cms-clot-video-vimeo">
```
CthulhuDev, line 2: wrong class name

patzick: thanks! :)

packages/cli/src/commands/cms.ts
```diff
@@ -0,0 +1,62 @@
...
+      blocks: {},
+      elements: {},
+    };
+
+    // Read theme cms settings
+    await toolbox.resolveCms(mainCmsPath, aliases, cmsComponentsMap);
+
+    // Override CMS by plugins
+    const pluginsConfig = await toolbox.filesystem.readAsync(
+      join(`.shopware-pwa`, "pwa-bundles.json"),
```
mkucmus, line 27: could we decide which format of string literals is preferred in the entire project?
We've got inconsistent strategy to deal with strings;) 
sometimes its "" (double quoted), sometimes ' ' (single quoted), and sometimes it's a new  ` ` (template string). 
Even though the template strings are desired way to deal with dynamic parts and string concatonations, we still don't know if double of single quoted string is preferred when it comes to deal with simple strings ;) (the same thing with module imports). Please choose a good strategy, I believe it could be spreaded over entire project.

packages/cli/src/templates/cmsNameMapper.ejs
```diff
@@ -0,0 +1,31 @@
+// const componentsMap = <%- JSON.stringify(props.componentsMap) %>
+const componentsMap = {
```
mkucmus, line 2: what's .ejs format? :)

packages/default-theme/cms/CmsGenericBlock.vue
```diff
@@ -1,34 +1,34 @@
 <template>
-  <component :is="getComponent" :content="content" :style="slotStyles"/>
+  <component :is="getComponent" :content="content" :style="slotStyles" />
 </template>
 
 <script>
-import { getComponentBy } from "../cmsNameMapper";
+import { getCmsBlockComponent } from 'sw-cms/cmsNameMapper'
```
mkucmus, line 6: `...from @shopware-pwa/default-theme/sw-cms/cmsNameMapper` maybe?

packages/default-theme/cms/blocks/CmsBlockCategoryNavigation.vue
```diff
@@ -0,0 +1,44 @@
...
+      type: Object,
+      default: () => ({}),
+    },
+  },
+  computed: {
+    getSlots() {
+      return this.content.slots || []
+    },
+    getContent() {
+      return this.getSlots.length && this.getSlots[0]
```
mkucmus, line 28: why only the first one is picked up? is there any other slot available to get?

packages/default-theme/cms/blocks/CmsBlockDefault.vue
```diff
@@ -0,0 +1,40 @@
...
+  },
+  computed: {
+    getSlots() {
+      return this.content.slots || []
+    },
+    getContent() {
+      return this.getSlots.length && this.getSlots[0]
+    },
+  },
+}
```
mkucmus, line 31: this logic is exactly the same as we have in `packages/default-theme/cms/blocks/CmsBlockCategoryNavigation.vue` 
Can it be provided as a mixin or something else to avoid redundant code? ;) Or I don't get it :D

packages/default-theme/cms/elements/CmsElementProductCard.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <SwProductCard
+    :product="content.data.product"
```
mkucmus, line 3: we can also get the product object from computed property to avoid lot of dots here ;) or a helper is a good way to extract it? i don't know :+1:

packages/default-theme/cms/elements/CmsElementProductListing.vue
```diff
@@ -1,56 +1,60 @@
...
+        <div class="cms-element-product-listing__place-holder" />
       </div>
-      <SfHeading
-        v-else
-        title="No products found"
-        subtitle="let us look for them"
+      <SfPagination
+        class="cms-element-product-listing__pagination desktop-only"
+        :current="pagination.currentPage"
+        :total="Math.ceil(pagination.total / pagination.perPage)"
```
mkucmus, line 20: would be nice to have that logic moved somewhere else I think :)

packages/nuxt-module/src/cms.ts
```diff
@@ -0,0 +1,57 @@
+import path from "path";
```
mkucmus, line 1: :mage_man: magic :+1:

packages/default-theme/cms/CmsGenericBlock.vue
CthulhuDev: @mkucmus I think it gets resolved anyway via nuxt

packages/cli/src/commands/cms.ts
patzick: yeah, we should configure prettier for that :)
* template strings - strings with dynamic parts
* double-quote otherwise

currently, double quotes are set for example in ts files, but not in .vue files.
As an action point, I'm creating an issue

patzick: https://github.com/DivanteLtd/shopware-pwa/issues/634

packages/cli/src/templates/cmsNameMapper.ejs
patzick: it's used for templates https://ejs.co/
this template generates us dynamically proper components resolving :)

packages/default-theme/cms/CmsGenericBlock.vue
patzick: 'sw-cms/cmsNameMapper' file is dynamically created in `.shopware-pwa/sw-cms/cmsNameMapper` path. Thanks to this we can easily resolve CMSes.
Folder `sw-cms` does not exist in default-theme nor will exist in any theme

packages/default-theme/cms/blocks/CmsBlockCategoryNavigation.vue
patzick: Every block is treated individually. This particular block has only one slot in its definition so there is no possibility for him to have more slots :)

definition:
https://github.com/shopware/platform/blob/master/src/Administration/Resources/app/administration/src/module/sw-cms/blocks/sidebar/category-navigation/index.js

packages/default-theme/cms/blocks/CmsBlockDefault.vue
patzick: good point, created issue for that :) https://github.com/DivanteLtd/shopware-pwa/issues/636

packages/default-theme/cms/elements/CmsElementProductCard.vue
patzick: will be covered with issue https://github.com/DivanteLtd/shopware-pwa/issues/636

packages/default-theme/cms/elements/CmsElementProductListing.vue
patzick: sure, but with another PR :)

packages/nuxt-module/src/cms.ts
patzick: https://nodejs.org/api/path.html (:

packages/default-theme/cms/sections/CmsSectionSidebar.vue
```diff
@@ -0,0 +1,75 @@
...
+    getBlocks() {
+      return this.content.blocks || []
+    },
+    getSidebarBlocks() {
+      return this.getBlocks.filter(
+        (block) => block.sectionPosition === 'sidebar'
+      )
+    },
+    getMainBlocks() {
+      return this.getBlocks.filter((block) => block.sectionPosition === 'main')
```
elkmod, line 44: Code style, compare to line 39-41.

patzick: prettier does that automatically :)
one line is clear to read,  but seems like in 39 it was too long

PR #631, opened 2020-04-21
feat(client): add service to interact with plugins

mkucmus:
### Checklist


mkucmus:
@patzick consider using it that way without providing a full scope of get, post, patch, etc. methods. 
At this point, it's easy to read, it's consistent, and every pwa plugin could use it like this: https://github.com/mkucmus/SwagPayPalPwa/blob/c3fe7b0901fc5478e7f5cf1773fdf9e7b6fb001d/src/Resources/app/pwa/main.vue#L17

PR #663, opened 2020-04-28
fix(cli): windows cli run

patzick:
## Changes

* remove optional chaining from CLI as it's not working in windows and was used only in two lines
* windows resolve of cms paths

### Checklist

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -100,12 +100,12 @@ module.exports = {
       await testPromise();
       // console.error("respo4", respo4.data);
     } catch (e) {
-      if (e?.response?.status === 401) {
+      if (e && e.response && e.response.status === 401) {
         error(
           `You provided bad cridentials for your shopware instance: ${inputParameters.shopwareEndpoint} - plugins will not be added`
         );
       } else {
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/generate.ts
```diff
@@ -100,12 +100,12 @@ module.exports = {
       await testPromise();
       // console.error("respo4", respo4.data);
     } catch (e) {
-      if (e?.response?.status === 401) {
+      if (e && e.response && e.response.status === 401) {
```
mkucmus, line 103: say goodbye, optional chaining :wave: :(

packages/cli/src/extensions/cms-extensions.ts
```diff
@@ -19,7 +19,7 @@ module.exports = (toolbox: GluegunToolbox) => {
         const path = resolve(join(directoryPath, sectionKey, value));
         const configPathExists = toolbox.filesystem.exists(`${path}.vue`);
         if (configPathExists) {
-          aliases[alias] = path;
+          aliases[alias] = path.replace(/\\/g, "/");
```
mkucmus, line 22: windows, we are not scared of you anymore ;)

PR #695, opened 2020-05-05
Feat plugin slots

### LINKED ISSUE #617: Implement named plugin slots
**Blocked by**
* #582 Proposition for named plugin slots

**Acceptance Criteria**
- [ ] Refactor theme to raise its modularity
- [ ] Implement slots in theme

patzick:
## Changes

closes #617 

Docs:
https://shopware-pwa-docs-git-feat-plugin-slots.shopware-pwa.now.sh/guide/plugins.html

Provides a way to:
* implement plugin locally in project
* override plugins
* extend plugins (multiple in a single slot)
* slots in theme

### Checklist


docs/guide/plugins.md
```diff
@@ -0,0 +1,202 @@
...
+    <slot />
+  </div>
+</template>
+```
+
+So this allows our installed plugin to show up and present itself.
+
+![local-plugin-3](./../assets/local-plugin-3.png)
+
+In this case, it shows a random plugin bud it could be as well another promotion plugin. You always decide where to show slot with another plugin if exists. There are a lot more use cases. Like social login button - you're doing plugin for Facebook, but others may do plugins for Google, GitHub etc.
```
elkmod, line 202: "but" instead of "bud"

docs/guide/plugins.md
```diff
@@ -0,0 +1,202 @@
...
+so we achieved this:
+
+![local-plugin-2](./../assets/local-plugin-2.png)
+
+### Nesting plugins
+
+Okay, but what if we have two or more plugins for our `top-header-after` slot?
+Always the newest installed plugin in shopware instance is on top. But local plugins are over them. The easiest way is to picture that. On shopware instance, we have our HelloCody plugin, which displays random quotes. It injects into `top-header-after` slot as well. So we overrode that.
+
+:::warning Always provide a slot for other plugins
```
elkmod, line 176: How about the ordering of plugins? If there is no determined order, maybe point out, that the point of truth is in fact the installation date in SW?

docs/guide/plugins.md
```diff
@@ -0,0 +1,202 @@
...
+
+It shows then a button in the footer, which allows us to turn on Dev mode, and see all plugin slots in theme.
+
+![turn-on-dev-mode](./../assets/turn-on-dev-mode.png)
+
+it will show us available slots on the theme, and you can click on any of them to see its name in the console
+
+![theme-slots](./../assets/theme-slots.png)
+
+### Local plugin
```
elkmod, line 21: Maybe we can also build the bridge to the Shopware instance plugins here, saying that `shopware-pwa init` will take the resources from each plugins `Resources/app/pwa` directory and place them within sw-plugins, so people will understand why they end up in this directory. Obviously in your example, they are writing the plugins directly into `sw-plugins` but it should be clear, that this is also the connection point to SW-Plugins.

docs/guide/plugins.md
```diff
@@ -0,0 +1,202 @@
...
+export default {
+  name: "CoolPromotionPlugin",
+  data() {
+    return {
+      product: null,
+    };
+  },
+  async mounted() {
+    // let's get the cheapest product
+    const productsResult = await getProducts({
```
elkmod, line 128: Maybe `{ productResult } = await...`?

packages/cli/src/templates/plugins/SwPluginEmptySlot.vue
```diff
@@ -0,0 +1,3 @@
+<template>
+  <div></div>
```
elkmod, line 2: Would really appreciate empty slots to resolve to empty markup, is that an issue? Seems like unnecessary overhead.

packages/default-theme/components/SwFooter.vue
```diff
@@ -1,40 +1,44 @@
 <template>
   <div class="sw-footer">
     <slot class="sw-footer__content" name="content" v-bind="column">
-        <div class="content sw-footer__signature">
-          Made with ❤️ by shopware AG & Vue Storefront
-        </div>
+      <div class="content sw-footer__signature">
+        <SwPluginSlot name="footer-content-before" />
+
+        Made with ❤️ by shopware AG & Vue Storefront
```
elkmod, line 7: It is more of a general thing - but I'd like to put it up for discussion. How about if these kinds of elements were itself put into plugin slots.

Few developers will just extend the footer, but instead want to replace the entire content of it. So for now they would have to replace the entire SwFooter component just to get rid of the content.

Maybe something like:

```suggestion
        <SwPluginSlot name="footer-content" default="Made with ❤️ by shopware AG & Vue Storefront" />
```

I am not sure, this lets us easily override or remove the content, but when it comes to deeper nesting this method doesn't work either. But do you get my concern?

packages/default-theme/components/SwProductDetails.vue
```diff
@@ -10,9 +10,11 @@
         :price="getPrice | price"
       />
     </div>
+    <SwPluginSlot name="product-page-description-before" />
     <p class="product-details__description desktop-only">
       {{ description }}
```
elkmod, line 15: Similar thing here - let's say you have a custom description field you'd like to display. How would you do that (without using CSS of course).

docs/guide/plugins.md
```diff
@@ -0,0 +1,202 @@
...
+- slot, in which we want to inject
+- vue file, which should be displayed there
+
+In the single plugin, we may inject in multiple plugin slots. Let's inject our `main.vue` file into `top-header-after` slot.
+
+```json{4,5}
+{
+  "slots": [
+    {
+      "name": "top-header-after",
```
elkmod, line 59: Can one plugin inject multiple components into the same slot or would they have to be composited into a single one?

docs/guide/plugins.md
patzick: the line above is about the order
`Always the newest installed plugin in shopware instance is on top. But local plugins are over them. `
maybe somehow change that?

docs/guide/plugins.md
patzick: we cannot do that, getProducts returns `SearchResult<Product[]>` so 
```
{
 total: number;
 data: Product[]
}
```
we could do
```
const {data} = await ...
```
but I'm not sure we should do this in example :)

packages/default-theme/components/SwFooter.vue
patzick: yes, that's a good point! we can do something like this:
``W
<SwPluginSlot name="footer-content">
  Made with ❤️ by shopware AG & Vue Storefront
</SwPluginSlot>
```

then the plugin may get rid of it by not putting `<slot />` inside its body or place it somewhere. Is that okay? :)

File: docs/guide/plugins.md (hunk 8 above)
Comment (patzick) [reply docs/guide/plugins.md:59]: no, it always should be a single component at the end as it some stage we'll add also context prop to API, so, for example, some slots will get current category or product as prop.

Why would you want to add two components? Could you show some use case for this?

File: docs/guide/plugins.md (hunk 1 above)
Comment (patzick) [reply docs/guide/plugins.md]: https://github.com/DivanteLtd/shopware-pwa/pull/695/commits/bf2f4561abada74ea64364c83ecc4ec8028de7d8

File: docs/guide/plugins.md (hunk 2 above)
Comment (patzick) [reply docs/guide/plugins.md:178]: https://github.com/DivanteLtd/shopware-pwa/pull/695/commits/bf2f4561abada74ea64364c83ecc4ec8028de7d8 is that any better? :)

File: docs/guide/plugins.md (hunk 3 above)
Comment (patzick) [reply docs/guide/plugins.md:23]: https://github.com/DivanteLtd/shopware-pwa/pull/695/commits/bf2f4561abada74ea64364c83ecc4ec8028de7d8

Comment (patzick) [reply docs/guide/plugins.md:23]: is that okay?

File: packages/cli/src/templates/plugins/SwPluginEmptySlot.vue (hunk 5 above)
Comment (patzick) [reply packages/cli/src/templates/plugins/SwPluginEmptySlot.vue]: removed this component :)

File: packages/default-theme/components/SwFooter.vue (hunk 6 above)
Comment (patzick) [reply packages/default-theme/components/SwFooter.vue]: https://github.com/DivanteLtd/shopware-pwa/pull/695/commits/b18fb19978927564ad15854dab22dbc2dd3e5303#diff-cfb20b6a177387e46c01724b35356cf7 great input Dominic! Works like a charm and allows both override and extend it :)

File: packages/default-theme/components/SwProductDetails.vue (hunk 7 above)
Comment (patzick) [reply packages/default-theme/components/SwProductDetails.vue]: https://github.com/DivanteLtd/shopware-pwa/pull/695/commits/b18fb19978927564ad15854dab22dbc2dd3e5303#diff-92fbafed2edd6afbfc8768ae9dbf3bf9 done as well :)

File: docs/guide/plugins.md (hunk 4 above)
Comment (elkmod) [reply docs/guide/plugins.md:130]: That's what I meant, my mistake! :) I thought it could be more concise, but I'll leave that up to you.

File: docs/guide/plugins.md (hunk 8 above)
Comment (elkmod) [reply docs/guide/plugins.md:61]: I'm not sure, that was just a question that came up to me. I think it's always possible to create a "container" component that just renders both of the components you want to add instead of adding them individually.

File: docs/guide/plugins.md (hunk 3 above)
Comment (elkmod) [reply docs/guide/plugins.md:23]: Yup, should do it!

File: docs/guide/plugins.md (hunk 2 above)
Comment (elkmod) [reply docs/guide/plugins.md:178]: So I assume the order for multiple plugins that use the same slot is determined by the order in `local-plugins.json`? It's an edge case but at some point, people will ask and we might just include it as well.

File: docs/guide/plugins.md (hunk 4 above)
Comment (mkucmus) [reply docs/guide/plugins.md:130]: I know it's only an example piece of code but we should always remember to catch the exceptions that may be thrown by the client :smile: otherwise bad things happen in the runtime :) It's related to upcoming notification system (not ready in UI) for asynchronous requests... just saying here, but maybe it's not the best place to talk about.

File: packages/cli/src/commands/plugins.ts [hunk 9]
```diff
@@ -0,0 +1,112 @@
...
+        shopwareUsernameQuestion,
+        shopwarePasswordQuestion,
+        devModeQuestion,
+      ]);
+      Object.assign(inputParameters, answers);
+    }
+
+    const allowDevMode = !!inputParameters.devMode;
+
+    await toolbox.loadPluginsAssets();
```
mkucmus, line 46: in the future, we should prevent some possible timeouts/errors during the downloading and extracting/processing the cms components from SW6. I believe it can be even 20 plugins maybe more with a considerable size.

patzick: sure, for now, it has loading indicator with CLI, we'll add something here when we will face real issue to test :)

docs/guide/plugins.md
patzick: good point, added that as well :)
https://github.com/DivanteLtd/shopware-pwa/pull/695/commits/a5dbcaca65ad5f0bc70b37246b3026ddf86b6c0d

docs/guide/plugins.md
patzick: I believe it's something we can consider in future when we'll have a use case for that, I see some dangerous area here:
* how to display these blocks? In a row, column?
* Where to put slot content? In first? Second? Both?

When there is a single entry point for plugin developer he's able to determine all of that and get the desired result quickly.
I'm not saying we shouldn't think about something like this, just more like find some use cases for that and figure out how the best approach would be :) But it's another issue I believe.

packages/default-theme/components/SwFooter.vue
elkmod: Great job!

docs/guide/plugins.md
elkmod: I agree, let's live it at this for now.

PR #791, opened 2020-05-26
feat(default-theme): ability to override global styles and scss variables in single place 🎨 

### LINKED ISSUE #751: [BUG] Overriding Global CSS Variables
### Describe the bug
CSS global variables can not been overridden as expected! 

main.scss file has been created and added in nuxt.config.js in CSS section as follows: css: ['@/assets/scss/main.scss']! 
The problem is, new variables are less specific than the old ones.

patzick:
## Changes

closes #751 

Changes:
- theme has now `main.scss` file for global styles and `variables.scss` for variables
- `init` command creates a template for two files which are importing two mentioned above with ability to extend them
- imports in all components are changes from `~@storefront-ui/vue/styles` to local `@/assets/scss/variables` file which contains all variables from the theme, overridden with user preferences

### Checklist

### DIFF
--- a/packages/cli/src/extensions/nuxt-extension.ts
+++ b/packages/cli/src/extensions/nuxt-extension.ts
@@ -125,6 +125,19 @@ module.exports = (toolbox: GluegunToolbox) => {
       });
     }
 
+    // Add global SCSS file to config
+    const isGlobalScssFileAdded = await toolbox.patching.exists(
+      "nuxt.config.js",
+      `~assets/scss/main.scss`
+    );
+    if (!isGlobalScssFileAdded) {
+      await toolbox.patching.patch("nuxt.config.js", {
+        insert: `
+    '~assets/scss/main.scss',`,
+        after: "css: [",
+      });
+    }
+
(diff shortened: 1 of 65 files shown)


packages/cli/src/extensions/nuxt-extension.ts
```diff
@@ -169,6 +182,24 @@ module.exports = (toolbox: GluegunToolbox) => {
         props: {},
       });
     }
+
+    const isMainScssFileCreated = exists("./assets/scss/main.scss");
```
mkucmus, line 186: is the mechanism windows proof? I mean the backslashes :)

packages/cli/src/templates/variables.scss
```diff
@@ -0,0 +1,11 @@
+// we're importing here main style variables
+@import '~@shopware-pwa/default-theme/assets/scss/variables';
+
+// For available variables check SFUI docs - https://docs.storefrontui.io/customization.html#global-variables
```
mkucmus, line 4: <3

Issue #908, opened 2020-07-01
[BUG] [OR UNWANTED FEATURE] the count of the cart also shows discount items


mmeester:
### Describe the bug
Adding a promotional code to the cart results in an additional item in the count of the cart. 

Let's say you have 3 items in your cart and add a promo code to the cart it will result in a count of 4

PR #909, opened 2020-07-01
fix: only count products in cart #908

mmeester:
## Changes
closes: #908 

### Checklist


patzick:
hey @mmeester thanks, good point with that issue, we just aim to have 100% tests coverage, but here tests are failing. Could you fix that? :)
you need to run `yarn test` and it will show you tests result and coverage

PR #925, opened 2020-07-03
feat: extend plugin service with patch #924

### LINKED ISSUE #924: [FEATURE] Implement PATCH invoke in shopware-6 client for plugin usage
### Description
At the moment only, GET and POST are available for custom calls, but for some plugins (like the stripe plugin) I need to call a PATCH call

mmeester:
## Changes
closes #924 

### Checklist

### DIFF
--- a/packages/shopware-6-client/src/services/pluginService.ts
+++ b/packages/shopware-6-client/src/services/pluginService.ts
@@ -31,3 +31,22 @@ export function invokeGet(
 ): Promise<any> {
   return contextInstance.invoke.get(address);
 }
+
+/**
+ * Invoke custom PATCH request to shopware API. Mostly for plugins usage.
+ * You can skip domain and pass only endpoint ex. `/api/my/endpoint`
+ * @throws ClientApiError
+ * @beta
+ */
+export function invokePatch(
+  {
+    address,
+    payload,
+  }: {
+    address: string;
+    payload?: any;
+  },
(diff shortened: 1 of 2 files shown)


packages/shopware-6-client/src/services/pluginService.ts
```diff
@@ -31,3 +31,16 @@ export function invokeGet(
...
+ * Invoke custom PATCH request to shopware API. Mostly for plugins usage.
+ * You can skip domain and pass only endpoint ex. `/api/my/endpoint`
+ * @throws ClientApiError
+ * @beta
+ */
+export function invokePatch(
+  { address }: { address: string },
+  contextInstance: ShopwareApiInstance = defaultInstance
+): Promise<any> {
+  return contextInstance.invoke.patch(address);
```
mkucmus, line 45: Please handle the `payload` parameter as well - not only the address - as you predicted in your tests here:
```
address: "/some/patch/endpoint",
payload: {
   some: "payload",
},
```

mmeester: I also noticed it :), working on it, I was a bit roo fast with my pr :)

patzick:
@mmeester thank you for this input, closing with a deeper explanation here https://github.com/DivanteLtd/shopware-pwa/issues/924#issuecomment-654322450, as it will be provided with the next version :)

PR #935, opened 2020-07-07
[PoC] feat(deployment): add slack notification on preview deploy

mkucmus:
## Changes
uses https://github.com/marketplace/actions/webhook-slack-notify to inform about the deployed instances.

### Checklist

### DIFF
--- a/.github/workflows/deploy-preview-storefrontcloud.yml
+++ b/.github/workflows/deploy-preview-storefrontcloud.yml
@@ -82,7 +82,6 @@ jobs:
           SLACK_MESSAGE: 'Your PR was successfully deployed on StorefrontCloud :rocket:'
           SLACK_TITLE: PR Preview
           SLACK_USERNAME: StorefrontCloudBot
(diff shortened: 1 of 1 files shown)


patzick:
@mkucmus @akrajna what are we doing with this?
I believe it's not necessary as deployment comment has always the timestamp for previews, example:
image

so if checks passed, then you know that comment is refreshed

on slack - you don't have the context and you're not sure if you're always testing the latest PR (as they may be couple of them and with multiple commits) - easy to get lost in that

Give me your opinions :)

mkucmus:
We encountered some issues on k8s level like when one user is browsing specific preview instance, and some new PR is triggering the new preview deployment - the current preview is just vanishing and the new one takes its place, then the user who was working on the previous one, not it's getting 404's because only one preview instance can be deployed at the same time. This is why we need a notification that something described above has happened and we can expect some odd behaviour like 404 :)

akrajna:
is exactly as mkucmus wrote. It would be really helpful to know that someone is working on some preview so as not to be surprised  because only one preview instance can be deployed at the same time.

PR #1051, opened 2020-08-26
fix(api-client): add product to cart duplicate entries (#1051)

elkmod:
## Changes

If you put the same product into the cart one by one, it creates new items (since there's no `id` passed to the endpoint).

That change was introduced in the the store-api.

**Before:**

image

**After:**

image

### Checklist


patzick (approved):
👌

Issue #1097, opened 2020-09-15
[BUG] Shipping method isn't saved correctly on creating Guest Order


mmeester:
### Describe the bug
Hi all, I’m running into the issue that selecting a shippingmethod works until I create an order and the shippingmethod jumps back to the default shipping method:
- calling setShippingMethod({ id }) is making a PATCH request to the /context endpoint with the correct id
- the following GET request to /context shows the correct shipping method selected
- the cart is updated with the new selected shipping method
**So far, so good**
- Looking in the database in the cart table the shipping method hasn’t changed
- when calling the createOrder() method the order get’s created but with the wrong shippingmethod attached

Please have a look at this slack thread to see the whole issue: https://vuestorefront.slack.com/archives/CT7AQR695/p1599817895084300

mkucmus:
the issue is confirmed. patching the current context by providing `shippingMethodId` does not affect the current context -> the GET /context response's salesChannel.shippingMethodId does not match the provided one, thus the placed order has a different one. 

patching the context with the specific method id responds correctly and looks like everything went right:
![image](https://user-images.githubusercontent.com/5596960/106204260-fe1d2180-61bc-11eb-8e89-a28936af6c2f.png)
context token is the same

![image](https://user-images.githubusercontent.com/5596960/106204418-37559180-61bd-11eb-8b9a-d3b4c9f6ceb0.png)
but the current context's shipping method has not changed.

@elkmod I think It looks like it was a core API issue.

mkucmus:
It might be cased due to an API bug, I've tested in in @canary version together with v6.4.0.0 RC1 and problem isn't present.

@mmeester  To avoid the issue I suggest upgrading both pwa and API to the latest versions and should be working well 👍🏻

PR #1098, opened 2020-09-15
fix: use CreateOrder() for both guest as logged in users (#1097)

mmeester:
## Changes
Solves #1097

### Checklist


mmeester:
This is not the solution 😔

PR #1101, opened 2020-09-16
fix(default-theme): hide error notifications when the list is empty

mkucmus:
## Changes
hide the encountered problems list above the specific form when the list is empty

### Checklist


patzick (approved):
slightly better if this would be a computed value, but can be corrected in future :)

PR #1127, opened 2020-09-22
chore(cli): Remove google analytics tracking from CLI

### LINKED ISSUE #1090: Google Analytics tracking in CLI
I just stumbled upon the following:
https://github.com/DivanteLtd/shopware-pwa/blob/e296bea203ea19d3d3a769e0625e0a954fa5b9ed/packages/cli/src/commands/override.ts#L61

Guys is it really necessary for you to track these kind of things silently to Google Analytics?

niklaswolf:
## Changes
Removed the Google Analytics tracking from CLI, as firstly it's really not nice to track without at least communicating it and secondly it's also questionable regarding GDPR (tracking without consent).
Personally I definitely do not want every override, build and init in my projects to be tracked.
closes #1090 

### Checklist


patzick:
I've prepared input for docs, but after internal discussion, we came to conclusion it's not worth it. So I'll paste it just here to clarify our motives. Personally I think it's a pity as stats of usage and product growth is a great team motivator. But if anyone feels that he's rights are violated in any way - no metrics then :)

So the input which will not reach the docs:
```
Like any other tool, we'd like to know that our project is in use. We need metrics to see project growth and potential areas to improve. However, we're not collecting any personalized data, session tracking or usage statistics by individual entities. Every CLI invocation is not connected to the other one.

What do we collect by Google Analytics is:

- number of `init` invocations - in order to see usage and how many times the project was initialized
- `init` command time invocation - we collect data of that method timing in order to have the ability to improve this timing with proper metrics as a reference
- `components overwrite` - in order to see which components are most often overwritten, it helps us to understand the struggles and provide help with prioritization
- running `dev` command invocation (no additional params) - to see approximately how many developers are using the tool in time, to be able to show some cool stats in future when the project will grow
- running `build` command (no additional params) - to see how many times Shopware PWA was built at all

We're considering to add additional methods for errors tracking or methods invocations timing, to see when CLI/nuxt module is causing some problems and where should we focus our priorities. Google Analytics groups invocations by the Country, but IPs are not stored or tracked by us in any way.  
That's it. We **do not** collecting any project names, structures, aby data that would allow us to connect statistics with any specific project/person (no sessions, every invocation is treated as "new user") as it's not our meaning.
```

Cheers!

elkmod:
Hi Niklas,

thanks for your pull request.

Our assumption was, that everybody would benefit from these statistics as we use them to track down potential bottlenecks within our application, such as:

* Most overridden components (indication for improvements)
* Init runtime information (to see how it behaves on different environments)

We'd like to point out, that we don't collect any data that will allow us to identify any project, customer, user related data - every tracking event is handled separately and doesn't allow us to draw any cross-references.

However, we agree with your note regarding user consent, even though our license explicitly says, that our product is open source, free of charge and comes without any liability for potential data infringements, we've decided to follow your request.

It doesn't align with our values to silently collect data that people might not be pleased with. We hope you can follow our initial decision making and try to keep the dialog as open as possible!

Thanks for your contribution!

/cc @niklaswolf @andreasemer

niklaswolf:
@elkmod @patzick 
I totally do understand what your motivations were (and I assumed this even before your clarification :) ). I want to point out that I do and did not assume that you were using this for something bad :)

I like your statement 
> It doesn't align with our values to silently collect data that people might not be pleased with
and I think that corresponds with the general mindset of the whole open source community!

So thanks for treating this seriously and merging this so fast ❤️

Issue #1141, opened 2020-09-29
[BUG] local plugins are not loaded/registered properly


mkucmus:
### Describe the bug
![image](https://user-images.githubusercontent.com/5596960/94530988-2b23ec80-023c-11eb-90d2-b5b1c146cf03.png)

PR #1146, opened 2020-10-01
fix(cli): importing local plugins relative to root instead of src root

elkmod:
## Changes

Solves #1141


mkucmus (approved):
works good :+1:

PR #1243, opened 2020-11-18
fix(helpers): extract option name from translated object

mkucmus:
## Changes

Always try to use the translated label/name of an option. If it's not possible, use the entity's name directly. 

### Checklist

### DIFF
--- a/packages/helpers/src/product/getProductOptions.ts
+++ b/packages/helpers/src/product/getProductOptions.ts
@@ -23,13 +23,14 @@ export function getProductOptions({
     }
 
     for (let option of variant.options) {
-      if (option.group?.name) {
-        if (!typeOptions.hasOwnProperty(option.group.name)) {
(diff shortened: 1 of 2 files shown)


Malwurf (approved):
Yeah! Now we know blau color :blue_heart:

packages/helpers/src/product/getProductOptions.ts
```diff
@@ -23,13 +23,14 @@ export function getProductOptions({
     }
 
     for (let option of variant.options) {
-      if (option.group?.name) {
-        if (!typeOptions.hasOwnProperty(option.group.name)) {
-          typeOptions[option.group.name] = [];
+      const groupName = option.group?.translated?.name || option.group?.name;
```
elkmod, line 26: I think we've got a couple of places left in our code, where we go straight for the property instead of the translated.property. We should, in general, always go for the translated field and never use the original field of any entity.

Issue #1247, opened 2020-11-19
[FEATURE] Allow disconnect of existing interceptors


niklaswolf:
### Description
For some interceptors anonymous functions are used as callbacks to the `intercept` function. The problem with this approach is, that it's impossible to disconnect them afterwards, as the `disconnect` function (see [here](https://github.com/DivanteLtd/shopware-pwa/blob/872782191be055ab953a2ec9ec0fb42ea0762004/packages/composables/src/logic/useIntercept.ts#L88)) expects the callback function used as a second argument and there is no possibility to get the reference of the anonymous function used in `intercept`.

Example for this usecase are the notifications registered in [notifications.js](https://github.com/DivanteLtd/shopware-pwa/blob/master/packages/default-theme/src/plugins/notifications.js).
Right now it's impossible to disconnect for example the add-to-cart notification which should definitely be possible.

A good approach could be to use named functions for the interceptor callbacks and export them. Like this one could import them for example in another plugin and disconnect single interceptor-functions.

### Acceptance criteria
- [ ] it is possible to disconnect single interceptors for example in another nuxt plugin

mkucmus:
Great! I agree we need to provide a more extensible API of useInterceptor composable... unfortunately when it comes to comparing anonymous function it does .toString on the method so practically impossible to find and disable such function.

niklaswolf:
I think this can be closed now since the code was merged :)

mkucmus:
> I think this can be closed now since the code was merged :)

you are right ;) in further PRs, please add the "closes: #xxx" sentence in the comment, and will be closed after the merge atumatically :+1:

niklaswolf:
I know, in all other PRs I've done it, apparently I forgot in this one.

PR #1249, opened 2020-11-20
feat: allow disconnect of existing interceptors

niklaswolf:
## Changes
This is a draft implementation to allow the disconnection of existing interceptors as described in #1247.
It goes beyond just a simple refactor, because I had to tackle two problems:

- You need access to the exported interceptor-functions, so I had to place them in a package. I decided to go with composables over helpers, because I need access to another composable in the notification-functions and did not have any dependencies from helpers to composables. But I really don't know if that's the place where it's supposed to be...
- You need access to the root context in the interceptor-functions (e.g. for i18n). So i decided to pass it automatically to every callback function as a parameter during the broadcast of the event 

## Testing
- create a new nuxt-plugin in test-project, e.g. `customPlugin.js`
```
import {INTERCEPTOR_KEYS, useIntercept} from "@shopware-pwa/composables";
import {addToCartNotification} from "@/logic/notifications";

export default async ({ app }) => {
  const { disconnect } = useIntercept(app)
  disconnect(INTERCEPTOR_KEYS.ADD_TO_CART, addToCartNotification)
}
```
- add the plugin to `nuxt.config.js`
```
plugins: [
    {src: '~/plugins/customPlugin.js', mode: 'client'}
  ],
```
- start the test project and add a product into the cart: the notification will not be shown.

### Checklist


niklaswolf:
@mkucmus maybe you can have a look, not quite sure about naming etc.

packages/composables/src/interceptors/notifications.ts
```diff
@@ -0,0 +1,37 @@
+import {
```
niklaswolf, line 1: Maybe this should be placed in default-theme, not sure if that's the right place.

packages/composables/src/logic/useIntercept.ts
```diff
@@ -36,6 +36,10 @@ export const INTERCEPTOR_KEYS = {
   USER_LOGOUT: "onUserLogout",
 };
 
+export interface IInterceptorCallbackFunction {
```
niklaswolf, line 39: I'm not sure about this naming at all 😅

mkucmus: I like the Java-like interfaces (or even php naming convention ;) )

mkucmus: and please mark all exported things as @beta or @alpha... depending on the destination

packages/composables/src/interceptors/notifications.ts
mkucmus: Yes, definitely, If you find that's the most default-theme-related interceptor you can place it right there. 
It parses a response from API without any helper, so it would be too much bound to the specific implementation, so in the first iteration, the more safe place is the default-theme.

So let's stay with the theme-specific interceptors

niklaswolf:
@mkucmus I moved the notification functions to the default-theme and added version-tags where required.

mkucmus (approved):
Finally, I have tested it, and works well! sorry for the last conflict to be resolved (accept both in translations file) - it will be merged. Great job! :+1:

niklaswolf:
@mkucmus I resolved the conflict :)

Issue #1216, opened 2020-11-03
[BUG] Link to product in cart links to current page


niklaswolf:
### Describe the bug
Currently the product link of a cartItem (only name is linked) links to the current page and not to the product itself.

- add item to cart
- go to another product page
- open cart and check the link of the cart item --> it links to the current product page

### Additional context
- In`SwCartProduct` the prop `link` is not set for `SfCollectedProduct` which leed to an empty string in the link
- I think the cart-API currently does not give any product links

Issue #1194, opened 2020-10-20
[BUG] After update from 3.x.x to 4.2.0 some shopware helpers stopped working


Radek-Wawrzyk:
### Describe the bug
When I updated the shopwarePWA version from **3.x.x** to **4.2.0** few helpers started return **undefined** (**SwProductHeading** component) values:
- **getProductRegularPrice()**  returns **undefined**
- **getProductSpecialPrice()**  returns **undefined**

### Additional context
**Screenshot of computed properties (PWA shopware)**

![Screenshot from 2020-10-20 16-06-59](https://user-images.githubusercontent.com/25388307/96597636-573c0600-12ee-11eb-8cc2-4f7ad7124f8e.png)

Shopware6 PWA version: **^0.4.2**
Shopware6 Dashboard version: **6.3.0.1**

PR #1252, opened 2020-11-23
refactor: price helpers #1194

mkucmus:
## Changes

closes: #1194 
closes: #1216 

### Checklist


packages/default-theme/src/components/SwCartProduct.vue
```diff
@@ -38,10 +41,32 @@ export default {
...
+          props.product.price.listPrice.price) ||
+        props.product.price.unitPrice
+    )
+    const specialPrice = computed(
+      () => props.product.price.listPrice && props.product.price.unitPrice
+    )
+
+    onMounted(async () => {
+      // async load of seoUrls only
+      const response = await getProduct(
```
niklaswolf, line 62: @mkucmus saw this by chance: wouldn't it be better to move this to SwCart, make a batch request (listing endpoint?) there and give the URL as a prop to SwCartProduct? Imagine having 10+ items in cart, this would mean lots of requests...

mkucmus: @niklaswolf thanks for watching :muscle: Yeah, It would be way efficient. It's the draft so it's not too late for little enhancements ;)

PR #1259, opened 2020-11-26
feat(client): possibility to pass the HTTP basic auth in project settings

mkucmus:
image

### Checklist


mkucmus:
It was rejected because HTTP basic auth is not safe anyway. 
credentials can be viewed after the simple base64 decode operation.

PR #1269, opened 2020-12-02
fix(default-theme): hide language switcher and currency switcher if only one available (bottom navigation)

jnuricumbo:
Language and currency switcher only visible in bottom navigation if more than one entry exist


mkucmus (approved):
:clap: nice :+1: thanks for that!

PR #1282, opened 2020-12-09
feat(default-theme): dynamic visibility of country state selection

### LINKED ISSUE #1263: [FEATURE] Show and validate state-selection based on configuration from API
### Description
Currently in address forms (checkout, customer-registration) the state selection is always there and not dependent on the configuration values from the API (`displayStateInRegistration`, `forceStateInRegistration`).
These values already exist in the application context under `salesChannel.countries[x].displayStateInRegistration` because the call to `/sales-channel-api/v3/country` already is in use. This endpoint also gives the before mentioned two fields for every country.

### Acceptance criteria

Address forms should respect this configuration:

- only show state field when `displayStateInRegistration: true` for the current country
- only require state field when `forceStateInRegistration: true` for the current country

niklaswolf:
## Changes
Outline of this feature is described in #1263. closes #1263 
This now works for all address forms:
- Checkout BillingAddressGuestForm
- Checkout ShippingAddressGuestForm
- Customer-Registration: SwAddressForm

what changed from the outline:
- currently there is no possibility in the shopware-admin to set `displayStateInRegistration`, only `forceStateInRegistration` has a switch. So the current implementation in this PR also looks for `forceStateInRegistration` when detecting if the country state should be displayed. This has to be changed when there is a possibility in the shopware-admin to set `displayStateInRegistration`

## Test
- set the value of "Force State in registration" in the Shopware Admin for some countries (Settings -> Countries -> Germany)
Bildschirmfoto 2020-12-09 um 11 25 31
- in shopware-pwa go to a address form listed above
- when selecting a country that has `forceStateInRegistration` set to true, the country state field should be displayed and be required!
- when selecting a country that has `forceStateInRegistration` set to false, the country state field should not be displayed

## Follow up tasks
- country state field should be a select with the values for the currently selected country. I did not get the states from the API, but apparently it used the wrong query: https://stackoverflow.com/questions/65201880/shopware-6-query-country-states-using-the-store-api/65213584#65213584
- address forms should be unified and use the same component, maybe `SwAddressForm`?

### Checklist


mkucmus (approved):
Works great, stable and it's ready :muscle: 
I just created a TODO-related issue and pasted it in useCountry composable :)

PR #1295, opened 2020-12-22
feat: domain handling

elkmod:
## Changes
shopware-pwa covers the "domains" logic coming from the admin panel settings. It's a single place responsible for: languages  (translations as well) and routing.

**It's more documented in Release notes & migration steps [LINK HERE].**

- there is another CLI command to synchronize the domains config
- useLanguages becomes deprecated -> useDomains is in charge now :)
- the pages structure does not require to use "_lang" param within the paths to get the right locale we are on
- /pwa/navigation endpoint is no longer used. built in /store-api/v{x}/navigation is being used since now


mkucmus (approved):
the main PR is enriched with the previous one containing most of logic was merged into it. It's good to merge it as well.

packages/default-theme/src/helpers/pages.js
```diff
@@ -1,5 +1,5 @@
 export const PAGE_CHECKOUT = "/checkout"
-export const PAGE_ACCOUNT = "/account/profile"
+export const PAGE_ACCOUNT = "/poland/test/account/profile"
```
elkmod, line 2: Still the error here

mkucmus: re-merge solved the issue ;)

packages/commons/interfaces/models/content/cms/CmsPage.ts
```diff
@@ -1,11 +1,51 @@
...
+ * @beta
+ */
+export interface PageResolverProductResult {
+  product: Partial<Product>;
+  aggregations: Aggregation[];
+  resourceType: PageType;
+  resourceIdentifier: string;
+  cannonicalPathInfo: string;
+  apiAlias: string;
+}
```
patzick, line 41: would be great to connect these interfaces, as ProductPage result is just another case of PageResolverResult. It should prevent from many type-related issues in future

packages/composables/src/appContext.ts
```diff
@@ -3,13 +3,28 @@ import { ShopwareApiInstance } from "@shopware-pwa/shopware-6-client";
 import { VueConstructor } from "vue";
 
 /**
- * Applicatoin Context for Shopware PWA. It's an extended Vue instance.
+ * @beta
+ */
+export interface Routing {
```
patzick, line 8: probably needs to be regenerated locally, as files above show warning related to this interface

packages/nuxt-module/plugins/domain.js
```diff
@@ -0,0 +1,92 @@
...
+    fallbackLocale: FALLBACK_LOCALE,
+    pwaHost: PWA_HOST,
+    // set current domain's configuration
+    setCurrentDomain: (domainData) => {
+      currentDomainData.value = domainData;
+    },
+    // get current domain's configuration
+    getCurrentDomain: () => {
+      return currentDomainData.value || routeDomain.value;
+    },
```
patzick, line 34: this one should be computed properties, not methods
```suggestion
    getCurrentDomain: computed(() => {
      return currentDomainData.value || routeDomain.value;
    }),
```

packages/composables/src/appContext.ts
```diff
@@ -3,13 +3,28 @@ import { ShopwareApiInstance } from "@shopware-pwa/shopware-6-client";
...
+  availableDomains: any;
+  fallbackDomain: string | undefined;
+  fallbackLocale: string | undefined;
+  pwaHost: string | undefined;
+  getCurrentDomain: () => string;
```
patzick, line 13:
```suggestion
  getCurrentDomain: ComputedRef<string>;
```

packages/default-theme/src/logic/useDomains.js
```diff
@@ -0,0 +1,81 @@
...
+
+export const useDomains = (rootContext) => {
+  const { router, routing, apiInstance } = getApplicationContext(
+    rootContext,
+    "useDomains"
+  )
+
+  const availableDomains = computed(() => routing.availableDomains || [])
+  const currentDomainId = computed(() => routing.getCurrentDomain().domainId)
+  const trimDomain = (url) => url.replace(routing.getCurrentDomain().url, "")
```
patzick, line 21: it's not reactive this way, added comments to make it computed property, so here:
```suggestion
  const currentDomainId = computed(() => routing.getCurrentDomain.value.domainId)
  const trimDomain = (url) => url.replace(routing.getCurrentDomain.value.url, "")
```

packages/default-theme/src/logic/useDomains.js
```diff
@@ -0,0 +1,81 @@
...
+  const getCurrentPathWithoutDomain = () =>
+    trimDomain(rootContext.$route.fullPath)
+
+  const isRouteStatic = () =>
+    !rootContext.$route.name.startsWith(PAGE_RESOLVER_ROUTE_PREFIX)
```
patzick, line 26: you shouldn't use rootContext directly (here or other places), that's why getApplicationContext does normalization, so we can use these values instead (i put `route` here but you can check correct value in interface
```suggestion
  const isRouteStatic = computed(() =>
    !route.name.startsWith(PAGE_RESOLVER_ROUTE_PREFIX))
```

packages/default-theme/src/logic/useLocales.js
```diff
@@ -0,0 +1,36 @@
...
+  const { i18n, router } = getApplicationContext(rootContext, "useLocales")
+
+  // TODO: consider using availabeDomains config to list the languages
+  const availableLanguages = computed(() => Object.values(languagesMap) || [])
+  // TODO: consider using locale code from current domains config
+  const currentLocale = computed(() => i18n.locale)
+
+  const changeLocale = async (localeCode) => {
+    // look for current domain
+    const domainForLocaleFound = rootContext.$routing.availableDomains.find(
```
patzick, line 15: same here - use values from getApplicationContext instead of rootContext directly

packages/default-theme/src/pages/account/addresses/add/_id.vue
```diff
@@ -1,4 +1,67 @@
+<template>
+  <div class="addresses-add">
+    <div v-if="address !== ''">
+      <SwAddressForm :address="address" />
+    </div>
```
patzick, line 5:
```suggestion
    <SwAddressForm v-if="address" :address="address" />
```

patzick (commented):
Great to see that huge progress here! Good job @elkmod @mkucmus!
I left some smaller suggestions I spotted. Keep it rolling 🏀  🙌

packages/composables/src/appContext.ts
mkucmus: right!

packages/default-theme/src/logic/useDomains.js
mkucmus: thanks!

PR #1296, opened 2020-12-29
refactor: product options

mkucmus:
### Checklist


packages/composables/src/internalHelpers/defaultApiParams.json
```diff
@@ -3,26 +3,30 @@
     "limit": 10,
     "associations": {
       "manufacturer": {},
-      "properties": {
+      "media": {},
+      "productReviews": {},
+      "options": {
```
elkmod, line 8: What do we need the `options` for?

packages/composables/src/internalHelpers/defaultApiParams.json
```diff
@@ -89,15 +93,16 @@
         "total",
         "aggregations"
       ],
-      "property_group": ["id", "translated", "options"],
+      "property_group": ["id", "translated", "name", "options"],
```
elkmod, line 96: Same here?

elkmod: We also don't need the name, if we have the `translated`.

packages/composables/src/internalHelpers/defaultApiParams.json
```diff
@@ -71,13 +74,14 @@
...
-        "optionIds"
+        "crossSellings"
       ],
       "product_media": ["media"],
       "media": ["thumbnails", "width", "height", "url"],
+      "media_thumbnail": ["url", "width", "height", "id"],
       "calculated_price": ["unitPrice", "quantity", "listPrice"],
       "product_group_option": ["name", "id", "group", "translated"],
-      "product_group": ["id", "name", "options", "translated"],
+      "product_group": ["id", "name", "translated"],
```
elkmod, line 84: No need for `name`.

mkucmus: right

packages/composables/src/internalHelpers/defaultApiParams.json
mkucmus: right

packages/composables/src/internalHelpers/defaultApiParams.json
mkucmus: I got rid of options association param, not everything is based on configurator + optionIds

packages/composables/src/internalHelpers/defaultApiParams.json
```diff
@@ -3,26 +3,25 @@
...
-      },
       "media": {},
       "productReviews": {},
-      "children": {
-        "associations": {
-          "options": {
-            "associations": {
-              "productOptions": {},
-              "group": {}
+        "crossSellings": {
```
elkmod, line 8: Wrong indentation.

This way it also fetches the crosssellings when listing pages are loaded? Will ultimately have a major performance impact. IMO we should remove this part from the initial page resolver request and do it asynchronously in the future. Wdyt?

packages/composables/src/logic/useProductConfigurator.ts
```diff
@@ -49,10 +49,30 @@ export const useProductConfigurator = (
...
+        (option: any) => option.id === optionId
+      );
+      return !!optionFound;
+    });
+
+    return group?.translated?.name;
+  };
+
+  // create a group -> optionId map
+  product.optionIds?.forEach((optionId) => {
```
elkmod, line 69: Why is this map required? Doesn't this return the same as `page.configurator` already contains? Looks redundant to me.

packages/composables/src/internalHelpers/defaultApiParams.json
```diff
@@ -67,17 +65,18 @@
         "parentId",
         "id",
         "translated",
-        "options",
+        "optionIds",
```
elkmod, line 68: Do we really need the `optionIds` when we have the `configurator` object?

packages/default-theme/src/components/SwProductDetails.vue
```diff
@@ -1,30 +1,31 @@
...
-        </div>
+    <div
+      v-if="product.optionIds && product.optionIds.length"
+      class="product-details__section"
+    >
+      <div v-for="config in getOptionGroups" :key="config.id">
+        <SwProductSelect
+          v-if="getSelectedOptions[config.translated.name]"
+          :value="getSelectedOptions[config.translated.name]"
+          :options="getProductOptions({ product: config })"
```
elkmod, line 17: Is this helper class still required or can we remove it and use plain `config` instead? Because it will already contain all the options for that group. => simpler to read-.

packages/helpers/src/product/getProductName.ts
```diff
@@ -1,17 +1,13 @@
...
   | string
   | null {
   if (!product) {
     return null;
   }
-  const variantLabel = getVariantOptionsLabel({ product: product });
-  return `${product.translated.name}${
-    variantLabel ? " - " + variantLabel : ""
-  }`;
+  return product.translated?.name || product.name;
```
elkmod, line 12: Screams to me like we could introduce a monad which models the fallback `[callee].translated.[key] => [callee].[key]` for translated strings. => less redundant code.

packages/composables/src/internalHelpers/defaultApiParams.json
mkucmus: right!

I've created an issue for that: https://github.com/vuestorefront/shopware-pwa/issues/1305

packages/helpers/src/product/getProductName.ts
mkucmus: related issue was created https://github.com/vuestorefront/shopware-pwa/issues/1302

elkmod (approved):

PR #1301, opened 2021-01-12
refactor: domain based routing & i18n

mkucmus:
## Changes

- get rid of _lang parameter globally and corresponding files (moved to the main pages path)
- provide nuxt plugin for domain management and usage
- enrich nuxt routes with domain-related data to be used later on - `build:extendRoutes` hook is used

TODO:
- language switcher:  languages source  :heavy_check_mark: 
- consider naming :heavy_check_mark: 
- fallback config - redirection if there is no matching domain  :heavy_check_mark: 
- docs - migration guide :heavy_check_mark: 
- CLI :heavy_check_mark: 
- CI / CD :heavy_check_mark: 
- Release notes 

### Checklist


packages/default-theme/src/helpers/pages.js
```diff
@@ -1,5 +1,5 @@
 export const PAGE_CHECKOUT = "/checkout"
-export const PAGE_ACCOUNT = "/account/profile"
+export const PAGE_ACCOUNT = "/poland/test/account/profile"
```
elkmod, line 2: Don't think this is correct?

mkucmus: you right :)  removed

packages/default-theme/src/middleware/auth.js
```diff
@@ -27,6 +27,6 @@ export default async function ({ route, redirect, app }) {
     isLoggedIn &&
     !isLoggedIn.value
   ) {
-    redirect(PAGE_LOGIN)
+    //redirect(PAGE_LOGIN)
```
niklaswolf, line 30: @mkucmus is this a TODO?

packages/nuxt-module/plugins/domain.js
```diff
@@ -0,0 +1,68 @@
...
+    // set current domain's configuration
+    setCurrentDomain: (domainData) => {
+      currentDomainData.value = domainData;
+    },
+    // get current domain's configuration
+    getCurrentDomain: () => {
+      return currentDomainData.value || domainFromRoute.value;
+    },
+    // get route for current domain
+    getRouteUrl: (path) =>
```
elkmod, line 35: As mentioned above, just `getUrl` to make it less verbose.

packages/nuxt-module/plugins/domain.js
```diff
@@ -0,0 +1,68 @@
...
+const FALLBACK_DOMAIN = "<%= options.fallbackDomain %>";
+const FALLBACK_LOCALE = "<%= options.fallbackLocale %>";
+const PWA_HOST = "<%= options.pwaHost %>";
+const domainsList = require("sw-plugins/domains");
+Vue.use(VueCompositionAPI);
+
+const currentDomainData = ref();
+// register domains based routing and configuration
+export default ({ app, route }, inject) => {
+  const routeDomain = computed(() => route.meta?.[0]?.url);
```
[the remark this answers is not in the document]
elkmod: Funny, because it switches names of the two, but I think it's more conclusive - wdyt?

packages/nuxt-module/plugins/domain.js
```diff
@@ -0,0 +1,68 @@
...
+    },
+    // get route for current domain
+    getRouteUrl: (path) =>
+      currentDomainData.value
+        ? `${currentDomainData.value.url}${path}`.replace(/^\/\/+/, "/")
+        : path,
+  };
+
+  // set the domain for current route
+  domainsRouting.setCurrentDomain(domainFromRoute.value);
```
elkmod, line 42: `setCurrentDomain(routeDomain.value)` to back what I wrote above.

packages/nuxt-module/plugins/domain.js
```diff
@@ -0,0 +1,68 @@
...
+Middleware.domainsRouting = function ({ isHMR, app, store, route }) {
+  if (isHMR) {
+    return;
+  }
+
+  // a route can have multiple metadata objects inside - find the one with domainId - it comes from compiled routes (domains nuxt-module)
+  const domainConfig =
+    Array.isArray(route.meta) && route.meta.find((data) => !!data.domainId);
+
+  if (domainConfig) {
```
elkmod, line 59: Back in the days I learned to write this as follows, but really just cosmetics - 

```js
if(!domainConfig) return // or whatever we want to check

[...]
```

packages/nuxt-module/plugins/i18n.js
```diff
@@ -1,10 +1,6 @@
 import Vue from "vue";
 import VueI18n from "vue-i18n";
-import Middleware from "./middleware";
-import languagesMap from "sw-plugins/languages";
-import domainsMap from "sw-plugins/domains";
-
-const FALLBACK_DOMAIN = "/"; // Replace with configuration
+import { useDomains } from "@shopware-pwa/composables";
```
elkmod, line 3: Is this import required?

packages/nuxt-module/src/domains.ts
```diff
@@ -0,0 +1,111 @@
...
+  DomainConfig,
+} from "./interfaces";
+import { NuxtRouteConfig } from "@nuxt/types/config/router";
+import jetpack from "fs-jetpack";
+
+export async function setupDomains(
+  moduleObject: NuxtModuleOptions,
+  shopwarePwaConfig: ShopwarePwaConfigFile | undefined
+) {
+  const domainsConfigExists = jetpack.exists(
```
elkmod, line 14:
```js
const domainConfigPath = path.join(
      moduleObject.options.rootDir,
      ".shopware-pwa",
      "sw-plugins",
      "domains.json"
    );

jetpack.exists(domainConfigPath);
```

packages/nuxt-module/src/domains.ts
```diff
@@ -0,0 +1,111 @@
...
+      "domains.json"
+    )
+  );
+  if (!domainsConfigExists) {
+    throw new Error(
+      "[shopware-pwa] Domains config is not initialized properly, please run 'shopware-pwa init'"
+    );
+  }
+
+  const domainsConfigFile = await jetpack.readAsync(
```
elkmod, line 28:
```js
const domainsConfigFile = await jetpack.readAsync(domainConfigPath);
```

packages/nuxt-module/src/domains.ts
```diff
@@ -0,0 +1,111 @@
...
+      name: `${parentRoute}_${route.path}_${domainId}`,
+      children:
+        route.children &&
+        appendChildrenWithUniqueName(route.children, route.name, domainId),
+    }));
+  };
+
+  /* istanbul ignore next */
+  const enrichRoutes = (routes: NuxtRouteConfig[]) => {
+    // reverse routes to set the global wildcard /* as a last one in the matching table
```
elkmod, line 60: 💡😏

packages/default-theme/src/middleware/auth.js
mkucmus: you're right! fixed :+1:

packages/composables/src/appContext.ts
```diff
@@ -2,6 +2,19 @@ import { getCurrentInstance } from "@vue/composition-api";
 import { ShopwareApiInstance } from "@shopware-pwa/shopware-6-client";
 import { VueConstructor } from "vue";
 
+/**
+ * @beta
+ */
+export interface DomainsRouting {
```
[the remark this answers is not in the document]
mkucmus: let be it :+1:

packages/default-theme/src/logic/useDomains.js
```diff
@@ -0,0 +1,26 @@
...
+import { getApplicationContext } from "@shopware-pwa/composables"
+
+export const useDomains = (rootContext) => {
+  const { i18n, router } = getApplicationContext(rootContext, "useDomains")
+
+  const availableLanguages = computed(() => Object.values(domains) || [])
+  const currentLocale = computed(() => i18n.locale)
+
+  const changeDomain = async (domainId) => {
+    console.warn(domainId)
```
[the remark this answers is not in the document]
mkucmus: right, removed

packages/default-theme/src/logic/useDomains.js
```diff
@@ -0,0 +1,26 @@
...
+    if (localeCode === i18n.locale) return
```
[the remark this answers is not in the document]
mkucmus: :+:

packages/nuxt-module/plugins/domain.js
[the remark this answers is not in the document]
mkucmus: I did it, except the route. I will stay as "routDomainUrl" because the route is currently reserved in that function :)

packages/nuxt-module/plugins/domain.js
mkucmus: the redirection to the default locale domain is being performed in case there is not domain found now

packages/nuxt-module/src/domains.ts
mkucmus: good point :)

packages/nuxt-module/plugins/domain.js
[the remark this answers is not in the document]
elkmod: Sounds good 👍  Totally missed that `route` was a parameter...

packages/composables/src/appContext.ts
```diff
@@ -2,6 +2,19 @@ import { getCurrentInstance } from "@vue/composition-api";
...
+  fallbackDomain: string | undefined;
+  fallbackLocale: string | undefined;
+  pwaHost: string | undefined;
+  getCurrentDomain: () => string;
+  setCurrentDomain: (domainData: any) => void;
+  getRouterUrl: (path: string) => string;
+}
+
 /**
  * Applicatoin Context for Shopware PWA. It's an extended Vue instance.
```
elkmod, line 19: Typo - "Application" might fix as well :D

packages/default-theme/src/components/SwBottomNavigation.vue
```diff
@@ -188,11 +188,11 @@ export default {
       } else this.toggleModal()
     },
     goToMyAccount() {
-      this.$router.push(this.$i18n.path(PAGE_ACCOUNT))
+      this.$router.push(this.$routing.getUrl(PAGE_ACCOUNT))
```
elkmod, line 191: I think this reads waaayy better 👍

packages/default-theme/src/logic/useDomains.js
```diff
@@ -0,0 +1,39 @@
...
+    if (!domainForLocaleFound) {
+      return
+    }
+
+    // remove domain
+    const cleanPath = rootContext.$route.fullPath.replace(
+      routing.getCurrentDomain().url,
+      ""
+    )
+    // build the new url based on current domain
```
elkmod, line 26: So right here we're stripping out the old domain from the path and replace it with the new domain. However, we can't be sure that this new path is a valid resource.

This is the place where we should check, what the correct path for the new domain is.

elkmod: For static pages the way given works, for page-resolver pages, we have to do the following:

```javascript
// POST http://localhost/store-api/v3/seo-url
// 
// Headers
//
// sw-language-id	2fbb5fe2e29a4d70aa5854ce7ce3e20b // New language id 

{
	"filter": [{
		"type": "equals",
		"field": "foreignKey",
		"value": "0096b08cf3524e29a6a4275988e45643"
	}],
	"includes": {
		"seo_url": ["seoPathInfo"]
	}
}
```

packages/default-theme/src/logic/useLocales.js
```diff
@@ -0,0 +1,36 @@
+import { computed } from "@vue/composition-api"
```
elkmod, line 1: Do you leave it for compatibility purposes?

elkmod (changes requested):

packages/default-theme/src/logic/useLocales.js
mkucmus: yes, I marked it as deprecated as well...

packages/cli/src/extensions/domains-extension.ts
```diff
@@ -48,7 +48,7 @@ module.exports = (toolbox: GluegunToolbox) => {
       }
     );
 
-    return fetchDomainsResponse.data.data;
+    return fetchDomainsResponse?.data?.data;
   };
 
   toolbox.domains.prepareDomainsMap = (domains, pwaHost: string) => {
```
mkucmus, line 54: @elkmod what about the domains starting with a hostname other than the passed `pwaHost`. maybe we should skip them and not include in the list of available domains? I think there will be a problem with the routing because our `availableDomains` would have a URLS with foreign domains - not matching the pwa's hostname.

elkmod: Yes, makes sense. We currently don't store any `pwa` flag for the domains. Let's just match `pwaHost` against the `domain.url` and remove the remaining domains.

packages/default-theme/src/logic/useDomains.js
```diff
@@ -1,34 +1,69 @@
...
+      try {
+        // find the correspoding URL for current page if it comes from page resolver - dynamically generated
+        const seoResponse = await getSeoUrls(
+          categoryId.value,
+          domain.languageId,
+          apiInstance
+        )
+        if (seoResponse.length && seoResponse[0].seoPathInfo) {
+          path += seoResponse[0].seoPathInfo
+        }
```
elkmod, line 45: Fallback when no SEO URL is found - go to either `/detail/${foreignKey}` or `/navigation/${foreignKey}`.

Important: As discussed, use constants instead of `/detail` and `/navigation`, because these are used in other places as well.

packages/default-theme/src/logic/useDomains.js
```diff
@@ -1,34 +1,69 @@
...
+
+  const isRouteStatic = () => !route.name.startsWith(PAGE_RESOLVER_ROUTE_PREFIX)
+  const getNewDomainUrl = async (domain) => {
+    let url = `${domain.url !== "/" ? `${domain.url}` : ""}`
+    let path = ""
+
+    if (isRouteStatic()) {
+      path += `/${getPathWithoutDomain()}`.replace(/\/\/+/, "/")
+    } else {
+      const { categoryId } = useCms(rootContext)
```
elkmod: Should apply to product detail pages and category pages equally.

elkmod (changes requested):
See comments regarding language / domain switch.

packages/default-theme/src/logic/useDomains.js
mkucmus: yes, I know that, categoryId points to resourceIndentifier, so it's working. I started to refactor the name of categoryId but it affects too many places... the PR would grow to 200 files, for real :)

packages/default-theme/src/logic/useDomains.js
mkucmus: the else statemend does that, it takes the URL from packages/helpers/src/cms/getCmsTechnicalPath.ts, there are a constant prefix either for navigation or for detail

packages/default-theme/src/logic/useDomains.js
elkmod: Okay, maybe a comment will do to have less confusion during future refactorings.

packages/default-theme/src/logic/useDomains.js
elkmod: I swear, that else part wasn't there when I commented.






Could be wrong though :D

packages/cli/src/extensions/domains-extension.ts
mkucmus: done :heavy_check_mark:

packages/default-theme/src/logic/useDomains.js
mkucmus: I left the appropriate comment, and also create an issue with TODO to refactor the useCms composable because of the misleading naming :)

elkmod (approved):

PR #1346, opened 2021-02-05
fix(default-theme): added translations for footer navigation and locales

elkmod:


mkucmus (approved):
:it: & :poland: here we come!

PR #1349, opened 2021-02-06
fix(default-theme): adding translated value to show more navigation item

elkmod:
This prevents a fatal error which was introduced with us rewriting the navigation to use translated values.


nhp:
THAT is really great. This helped a lot. Thanks for your super direct support @elkmod. That left an outstanding impression.

PR #1357, opened 2021-02-09
fix(default-theme): added shipping cost to order summary

elkmod:
## Changes

Shipping costs weren't correctly displayed in order summary.


packages/composables/src/hooks/useCart/index.ts
```diff
@@ -153,6 +154,11 @@ export const useCart = (rootContext: ApplicationVueContext): IUseCart => {
     return cartPrice || 0;
   });
 
+  const shippingTotal = computed(() => {
+    const shippingTotal = cart?.value?.deliveries[0]?.shippingCosts?.totalPrice;
```
elkmod, line 158: Is optional chaining allowed?

mkucmus: definitely you can, in composables the optional chaining is supported. in this case you can even use ....deliveries?.[0]? for arrays :+1: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#array_item_access_with_optional_chaining

mkucmus (approved):
great :1st_place_medal:

packages/composables/src/hooks/useCart/index.ts
elkmod: Alright, cool!

PR #1351, opened 2021-02-07
fix(default-theme): fix missing customer address right after user logs in 

jnuricumbo:
closes #1320

Problem: costumer context property is null right after user logs in, and therefore leads to crash on shipping checkout step.

As solution the context is refreshed after log in success.

### Checklist

- [ ] the [list of features](docs/guide/FEATURELIST.md) is updated (if it's a feature and it's relevant)
- [ *] I followed [contributing](https://github.com/DivanteLtd/shopware-pwa/blob/master/CONTRIBUTING.md) guidelines

PR #1365, opened 2021-02-14
fix(default-theme): missing customer address right after user registr…

jnuricumbo:
Same as #1351 but this time refreshing context after registration, otherwise it will crash on step two because there is no customer address.


jnuricumbo:
@mkucmus this fix the problem described in this PR, but I guess there is not the best fix, login method in useUser.ts should do the job.

PR #1385, opened 2021-03-01
refactor(default-theme): add vue extension to component imports

patzick:
## Changes

Following good practices component imports should have .vue extension:

> You should always use the .vue extension when importing Vue components, as omitting the extension creates numerous problems for static analysis tools. It was a mistake to allow it in vue-cli and we will likely disable it in the future.

source: https://twitter.com/youyuxi/status/1288859415878283264


mkucmus (approved):
welcome back! it's a huge first PR :rocket:

PR #1388, opened 2021-03-03
chore: update dependencies and node version in GH actions

patzick:
## Changes

* updated dependencies in packages
* use esbuild to build CLI
* switch GH actions to use Node 14
* moved husky hooks after upgrade


mkucmus (approved):
we are almost always up to date :-) :muscle:

PR #1395, opened 2021-03-14
refactor(default-theme): lazy load more components on initial page load

patzick:
## Changes

A little bit smaller initial bundle in default-theme.


mkucmus (approved):
being lazy is also very important! 😄

PR #1397, opened 2021-03-18
fix(default-theme): Remove unneeded parameter

develth:
## Changes

Removes an unneeded Parameter that appeared after #1385

### DIFF
--- a/packages/default-theme/src/components/listing/SwProductListingFilter.vue
+++ b/packages/default-theme/src/components/listing/SwProductListingFilter.vue
@@ -44,7 +44,7 @@ export default {
     getComponent() {
       try {
         return () => ({
-          component: import(`@/components/listing/types/${this.filterCode}`),
+          component: import(`@/components/listing/types/${this.filterCode}.vue`),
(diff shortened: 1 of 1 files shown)


packages/default-theme/src/components/listing/SwProductListingFilter.vue
```diff
@@ -52,7 +52,7 @@ export default {
       }
     },
     filterCode() {
-      return simplifyString(this.filter.label.toLowerCase(".vue"))
+      return simplifyString(this.filter.label.toLowerCase())
```
patzick, line 55: good point here! missed that in global replace
in 47. line instead of
```js
component: import(`@/components/listing/types/${this.filterCode}`),
```
should probably be
```js
component: import(`@/components/listing/types/${this.filterCode}.vue`),
```

could you add this there?

develth: Ok, added it!

patzick (approved):
👌  perfect! Thank you!

PR #1401, opened 2021-03-22
fix(default-theme): duplicated email field id on checkout

### LINKED ISSUE #1373: [BUG] Email Field Focus in Checkout Login
### Describe the bug

When users open the "Log in to your account" dialog on the checkout page, it's almost impossible to focus the email field.
Reason: There's another Email field that's covered by the login dialog and this one gets focus instead. For some reason, it only works if you click the bottom border of the login field. (See active field in the background on the screenshot below)

image

patzick:
## Changes

closes #1373 

Fixes the problem with input focus caused by inputs id duplication


mkucmus (approved):
let's duplicate you @patzick 😂  and we are stable in one month 👌🏻

PR #1402, opened 2021-03-24
feat: shared state composable to replace vuex and speed up page loading

patzick:
## Changes

**BREAKING CHANGE**: Vuex has been turned off and replaced with `useSharedState` composable.
If you want to persist your data between Server and Client, then setting `sharedRef` from `useSharedState` on the server side will cause that this ref is also filled in client side.

`useSharedState` is also adding very helpful method, which is invoked if Ref value is not setted.
Let's say we're loading Easter Eggs from the API, but we don't want to duplicate it both on SSR and CSR.

```ts
setup(root) {
  const {sharedRef, preloadRef} = useSharedState(root)

  const easterEggs = sharedRef('my-easter-eggs')

  preloadRef(easterEggs, async () => {
    easterEggs.value = await fetchEasterEggsFromAPI()
  })

  return {
    easterEggs
  }
}
```

We can assume this code is in some component on MyAccount page. We have two scenarios:

1. Client goes directly into `/my-account` page (or refreshes the page). Then `preloadRef` method will be invoked on server, value will be saved under key `my-easter-eggs` and restored on client side. So there will be no API call there.
2. Client goes into Home `/` and clicks MyAccount icon. Then we didn't had the chance to get this data on Server, so we'll call code from `preloadRef`.

This way we always call for API data, but we will not repeat API calls when we already have this value.


mkucmus (approved):
What a significant change!

What I found here is:
- no more vuex - what is cool 👍🏻 
- little workbox optimisation 🕐 
- resources loaded in ssr can be shared, but they are still in the current app context (prevent data leaks between contexts)
- `onServerPrefetch` hook is being used to get the resoures before page's load - 📉 
- jest tests are simplified 🍀

Issue #1403, opened 2021-03-27
[FEATURE] Extensible nuxt config from nuxt-module


patzick:
### Description

In order to avoid changing generated Nuxt config we should change the process of project initialisation.

1. Store nuxt.config.js template in `nuxt-module`. This way we always have config suited for specific module version
2. Change CLI process to generate something like

```js
// generated nuxt.config.js

const { extendNuxtConfig } = require("@shopware-pwa/nuxt-module")

export default extendNuxtConfig({
  head: {
    title: 'my project head title',
  }
}) 
```

### To be decided
- should nuxt config be on the theme as well? It could help with the theme settings

PR #1404, opened 2021-03-27
refactor(cli): improve CLI generation process, remove create-nuxt-app usage

patzick:
## Changes

- removed `create-nuxt-app` usage - we're using the generated template instead
- fetching released versions, we're showing specific versions in CLI 
image
- we can use specific version in cli `npx @shopware-pwa init --stage=0.6.0`

Thanks to this PR init process is much faster and more reliable.
Creating a new project in GH Actions went down from 1:40 to 0:59 ⚡ 

Additionally, later we can improve nuxt config, more in #1403


mkucmus (approved):
it's faster, it's clearer, has less dependencies... how can I reject it? 🚀

PR #1410, opened 2021-03-31
fix(cli): theme creation gitignore file

patzick:
## Changes

 Fixes problem with `create-theme` command causing error:
`Error: template not found /node_modules/@shopware-pwa/cli/build/build/templates//themes/.gitignore`

Cause - npm ignores .gitignore files by default while publishing packages, so this file could not be found.


mkucmus (approved):
🐛 fixed!

PR #1423, opened 2021-04-07
refactor(theme-base): move basic ui-independent components from default-theme to base-theme

patzick:
## Changes

* moved few ui-independent components to theme-base
* changed page resolver mapping by creating in `views` folder components with name as resolved page resource type


mkucmus (approved):
🆕 ✅

PR #1436, opened 2021-04-19
refactor: common default config across the packages, cli uses compatibility table

patzick:
## Changes

* CLI init recognizes API version and proposes proper default instance
* all the default settings across the packages are moved to commons and reused
* GH actions now use default shopware instances


mkucmus (approved):
🍯

PR #1447, opened 2021-04-26
feat: native TypeScript support in components, plugins, theme and project

patzick:
## Native TypeScript support

You can use TypeScript natively across the project, plugins, theme and logic.
Just add `<script lang="ts">` in your components to start using it there. All your files can be written in `.ts` files.

Quick example:
image

image

image

image


mkucmus (approved):
we gonna TS it out 💯

PR #1458, opened 2021-04-29
fix: default language settings

### LINKED ISSUE #1453: [BUG] Default language not correclty set
### Describe the bug

The default language is not correctly set if this is not english. This is happening in v.0.8.1, tested with a fresh instalation of shopware and shopware-pwa

### Steps to reproduce it

- Remove all domains if any and create only one like this
image

-  Update your domains using "yarn shopware-pwa domains"
- Start the shopware-pwa project, maybe clearing the cache is needed
- You will have german language for a second and then the language is set automatically to english, pay attention to the search field in the top bar.
- Setting defaultLanguageCode: 'de-DE' in config does not help.

patzick:
## Changes

closes #1453 
I'll do follow-up PR to master branch

### DIFF
--- a/packages/nuxt-module/plugins/i18n.js
+++ b/packages/nuxt-module/plugins/i18n.js
@@ -1,13 +1,18 @@
+import { useSharedState } from "@shopware-pwa/composables";
 import Vue from "vue";
 import VueI18n from "vue-i18n";
 
 Vue.use(VueI18n);
 
 export default ({ app }, inject) => {
+  const { sharedRef } = useSharedState(app);
+  const currentDomainData = sharedRef("sw-current-domain");
   // Set i18n instance on app
   // This way we can use it in middleware and pages asyncData/fetch
   const i18n = new VueI18n({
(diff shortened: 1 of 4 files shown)


packages/nuxt-module/plugins/i18n.js
```diff
@@ -1,13 +1,18 @@
...
 Vue.use(VueI18n);
 
 export default ({ app }, inject) => {
+  const { sharedRef } = useSharedState(app);
+  const currentDomainData = sharedRef("sw-current-domain");
   // Set i18n instance on app
   // This way we can use it in middleware and pages asyncData/fetch
   const i18n = new VueI18n({
-    locale: "en-GB",
+    locale:
```
mkucmus, line 13: I won't miss it anymore in CR ! :)

patzick: It wouldn't work, either way, the other plugin is changing this value dynamically :)

mkucmus (approved):

PR #1524, opened 2021-06-02
fix(nuxt-module): sort routes with localhost and port domains

patzick:
## Changes

Fix local routes when a domain like `:3000` is registered


mkucmus (approved):
good bye, edge case 👍🏻

PR #1557, opened 2021-06-16
fix(default-theme): salutation in user registration form

### LINKED ISSUE #1540: [BUG] Salutations missing in checkout
For billing address there is no salutation input, for shipping address the input is commented.
Also the salutation should be the first input of the form, in both cases.

Shopware-PWA 0.9.x
Shopware 6.4.1

https://github.com/vuestorefront/shopware-pwa/blob/382af7ac75d68cc6464aed30df541df886805029/packages/default-theme/src/components/forms/SwRegistrationForm.vue#L149

patzick:
## Changes

closes #1540 

Salutation id is not required, by default we take "not_specified"


mkucmus (approved):
salute 🖖🏻

PR #1558, opened 2021-06-16
fix(nuxt-module): revert devtools in non-production mode

patzick:
## Changes

fixes problem with unavailable devtools in non-production mode


mkucmus (approved):
👌🏻

PR #1563, opened 2021-06-16
fix(cli): adjusted project template prettier config

### LINKED ISSUE #1136: [BUG] Linting/formatting of vue components
### Describe the bug
Besides being set up as git hook, the formatting doesn't seem to be applied. All vue components in default-theme are not formatted according to prettierrc. For example there are double quotes everywhere besides single quotes being configured in prettierrc.

It's just a small issue, but it's a bit annoying because IF you have setup formatting/linting in your IDE every overridden component gets formatted according to prettierrc. That does make updating/comparing overridden components to their original more difficult than it needs to be, because you get all these formatting differences.

I like to propose, that all components get formatted according to prettierrc, alternatively the configuration of prettier should at least match the current formatting style.

### Additional context
https://github.com/DivanteLtd/shopware-pwa/blob/master/packages/default-theme/.prettierrc

niklaswolf:
## Changes
closes #1136

### Checklist


patzick (approved):
Good point, thank you! :)

PR #1577, opened 2021-06-28
WIP allow different domains for different locales

### LINKED ISSUE #1367: [FEATURE] allow different domains for different languages
### Description
@mkucmus as discussed, here is the issue :) 

Each of our clients has the following domain/i18n setup:
- domain.de --> german version
- domain.com --> english version

It should be possible to use different domains for the same sales-channel, but with different language settings.
Right now there seems to be some kind of filtering for the current pwaHost.
https://github.com/vuestorefront/shopware-pwa/blob/e04da66000f0fbf2bb2b78970073b1c796599ad8/packages/cli/src/extensions/domains-extension.ts#L57

This makes it impossible to link to different language versions dynamically (if these are on different domains), for example in the language switch.

Maybe there is an approach to make this configuration an array to be able to provide more than one domain for the current application?

cc @elkmod @mkucmus

niklaswolf:
## Changes
closes #1367


niklaswolf:
@patzick I think you now can have a first look at the code and design, as it seems like everything is working as intended... :)

atlantis-vsf:
Atlantis commands can't be run on fork pull requests. To enable, set --allow-fork-prs  or, to disable this message, set --silence-fork-pr-errors

PR #1592, opened 2021-07-09
feat: simplify page resolver logic, shared useCms loading and error state

patzick:
## Changes

Simplify flow of asyncData page resolver component. Logic is moved to middleware.


packages/theme-base/dist/pages/_.vue
```diff
@@ -1,49 +1,32 @@
...
 
-    const unwrappedPage = page && page.value ? page.value : searchResult;
-    const name =
-      unwrappedPage && unwrappedPage.cmsPage && unwrappedPage.cmsPage.name;
-    const cmsPage = unwrappedPage && unwrappedPage.cmsPage;
-
-    // redirect to the cannnical URL if current path does not match the canonical one
-    if (params.pathMatch && unwrappedPage && unwrappedPage.canonicalPathInfo) {
-      if (unwrappedPage.canonicalPathInfo !== params.pathMatch) {
-        redirect(301, app.$routing.getUrl(unwrappedPage.canonicalPathInfo));
```
mkucmus, line 39: shouldn't we move this logic to the page middleware as well?

patzick: do you mean this?
```js
const cmsPage = computed(() => page.value?.cmsPage);
const cmsPageName = computed(() => cmsPage.value?.name);
```
Cms name is only for DevTools, I believe we should pass the whole CMS page response into specific view (like in product page) so this computed won't be necessary anymore :)

mkucmus (approved):

packages/theme-base/dist/pages/_.vue
patzick: good point, thanks!

PR #1597, opened 2021-07-15
feat(cli): use current package version on project init

patzick:
## Changes

 Changes: 
- removed version picking, CLI uses currently installed version
- auto-generate env file based on 
- improved admin auth token fetching


mkucmus (approved):
🍺

Issue #781, opened 2020-05-25
Move OrderDetails logic from setup into separate composable


mkucmus:
Related component: packages/default-theme/components/SwOrderDetails.vue

Acceptance criteria:
- [ ] setup as tiny as possible
- [ ] the OrderDetails composable is created
- [ ] all tests are updated

niklaswolf:
It would also be great to have the order data broadcasted for interceptors to use it, e.g. for tracking.
The interceptor events are great for such a usecase, as you can access them from everywhere in the app, also plugins, which makes them really nice for a usecase like tracking, where you want to seperate the tracking-code as much as possible from your  app code.

Issue #717, opened 2020-05-08
[FEATURE] Implement order details component


patzick:
### Description

related #664 

### Acceptance criteria

- [x] implement order details component based on SFUI designs
- [x] component is used in my account and thank you page
- [x] in section status there is a payment button showing when payment is available (this is implemented, only place it inside order details)

PR #1603, opened 2021-07-23
feat(composables): useOrderDetails as order manager

mkucmus:
## Changes

closes: #717 
closes: #781 

TODO
- [x] add interceptors
- [x] add migration guide
- [x] add tests

### Checklist


packages/shopware-6-client/src/services/checkoutService.ts
```diff
@@ -39,3 +47,70 @@ export async function handlePayment(
...
+
+/**
+ * Get order details
+ *
+ * @throws ClientApiError
+ * @beta
+ */
+export async function getOrderDetails(
+  orderId: string,
+  params?: Object,
```
patzick, line 59: shouldn't it be searchParams? We need a type here

patzick: also, what would the additional params be? If we're fetching by id we don't need to pass any additional params I guess

patzick (commented):
Just one thing from me regarding `getOrderDetails`

patzick (approved):

PR #1611, opened 2021-08-04
fix(composables): memory leak in the page middleware

patzick:
## Changes

The issue raised to me by @niklaswolf 
We had a memory leak caused by the watcher invoked on the server-side in the page middleware. I traced that down and after this fix, I see no signs of that.
After we'll confirm that the problem is solved I'll backport it to 0.9 and 0.8 versions as it is a pretty important thing and the fix is not breaking anything.


atlantis-vsf:
```
Error: This repo is not allowlisted for Atlantis.
```

niklaswolf:
I can confirm that the leak is fixed :)

mkucmus (approved):
it's time to say "goodbye leaks" :-) 👍🏻

PR #1616, opened 2021-08-05
fix(client): pass parameters correctly within getProduct method

samuellembke:
## Changes

In the getProduct method is an attribute called params, wich is simply given to the request body as params.
This makes it that you cant give request config like associations, etc..
I believe this was by mistake, if so please accept the pull request.
If not, why? and what does params do, because I couldn't find any documentation or usage that would make sense.


mkucmus:
hi @samuellembke. thank you very much for the PR! Please update the tests to not lose code coverage 👍🏻

samuellembke:
@mkucmus What exactly do you mean by update the tests to not lose code coverage?
This is my first time I do a pull request for you guys

patzick (approved):
I fixed the test, thank you for the improvement!

PR #1619, opened 2021-08-09
fix(nuxt-module): ability to add new config key for defaults

### LINKED ISSUE #1514: [FEATURE] Access to plugin configurations (pwa-bundles.json)
### Description

Now the `pwa-bundles.json` gets generated when running `shopware-pwa plugins`. In the `pwa-bundles.json` some values i have to use in my code are defined. Think of apiKeys and some configuration values for my plugins. Is it a good idea to make this available also to the client in someway that we can blacklist sensitive keys or whitelist only the one we need?

### Example

The Array below exists in `pwa-bundles.json`:

```js
"shippingMethods": [
          "a4139c276ad24bf1ab51cebce3864345"
 ]`
```
This array consist of the id of the shippingMethod named pickup.

On the homepage i have a few dynamic buttons which will display a button to select the shippingMethod pickup. Now i need to hardcode the name of the shippingMethod to filter it out of the `shippingMethods` object of `useCheckout`. 

If i could use the array from `pwa-bundles.json`, it would be dynamic on buildtime .

### Acceptance criteria

- [blacklist sensitive values or whitelist non-sensitive values]
- [Make pwa-bundles.json available to use in code]

patzick:
## Changes

closes #1514 

Testing the scenario from the issue I saw that a new key is not created when the path is set.

### DIFF
--- a/packages/nuxt-module/__tests__/defaultsConfigBuilder.spec.ts
+++ b/packages/nuxt-module/__tests__/defaultsConfigBuilder.spec.ts
@@ -52,6 +52,17 @@ describe("nuxt-module - defaultsConfigBuilder", () => {
         associations: { media: {}, something: {} },
       });
     });
+
+    it("should add new key to config", () => {
+      const result = defaultsConfigBuilder()
+        .add("myNewKey", {
+          someConfig: 123,
+        })
+        .get();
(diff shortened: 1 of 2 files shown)


mkucmus (approved):
👍🏻

PR #1643, opened 2021-08-25
feat(cli): prefixes before notifications

mkucmus:
may be helpful in commands waterfall, especially doing within `init` process.

image


patzick (approved):
📓

PR #1653, opened 2021-09-06
fix(cli): downgrade md5-hex to use cjs in CLI

patzick:
## Changes

downgrade md5-hex to use cjs in CLI. The problem appeared in `snippets` command


mkucmus (approved):
good to see you, old 3.0.1 friend :) no esm here!

PR #1656, opened 2021-09-07
chore: remove root param from composables and old TODO marks

patzick:
## Changes

just a cleanup


mkucmus (approved):
using composables is easy like never before. well done! 🥇

PR #1661, opened 2021-09-13
fix(nuxt-module): language based on current route meta config

mkucmus:
## Changes

- decodes SHOPWARE_DOMAINS_ALLOW_LIST into array of strings (can be done as ejc for loop as well)
- allows to change the current config on CSR also.

TODO:
- check different ports
- check different domains over a proxy
- fix homepage redirection (no technical urls)

### Checklist


patzick (approved):
🎏

PR #1670, opened 2021-09-16
feat(helpers): getTranslatedProperty method

### LINKED ISSUE #1302: Screams to me like we could introduce a monad which models the fallback `[callee].translated.[key] => [callee].[key]` for translated strings. => less redundant code.
Screams to me like we could introduce a monad which models the fallback `[callee].translated.[key] => [callee].[key]` for translated strings. => less redundant code.

_Originally posted by @elkmod in https://github.com/vuestorefront/shopware-pwa/pull/1296#discussion_r554997587_

patzick:
## Changes

closes #1302 

Introduced `getTranslatedProperty` helper, which gets translated property from an object.
 provides autocompletion
image

and error checks for properties in object 
image


mkucmus (approved):
👍🏻

PR #1685, opened 2021-09-23
fix(theme-base): fix duplicate rendering of cms-page subtree

### LINKED ISSUE #1682: [Bug] duplicate creation/mount of views
### What happened?

Right now every view (`components/views/*`) gets created and rendered twice. This is a huge problem as the whole subtree i created and mounted twice which also potentially causes duplicate API-calls etc.

Expected is that the view-components and their subtree only get created once per page view.

### Steps To Reproduce
1) Go to a category page
2) Go to a product page
--> for example the cross-sells API-call is duplicated.
If you write a `console.log("created")` into the setup-method of for example `components/views/frontend.detail.page.vue`, it gets even more obvious since it appears twice in the console.

The root of problem is, that the pages-middleware is executing the search-method of useCms which then of course updates the page value. 
This causes the `getComponent` value in `pages/_.vue` to recompute which in turn causes the view-component to be created again. 
Shortly after the current page-instance gets unmounted and destroyed by the router and then a new instance of `_.vue` is created. This causes another creation and mount of the respective view-component.

As I see it, getComponent doesn't have to be a "real" computed value and can be executed in `setup` directly, because the page-value won't update during the lifecycle of its instance. The page-value should already be set on the creation of the _.vue-instance, because the middleware is always executed before route change.
Another option is to use `asyncData` on the page to essentially do the logic that is currently done in the pages-middleware. This would then not mess with the execution order and prevent unneccessary rerenders.

### Version

Canary (generated by CLI)

default-theme

### Environment

_No response_

### Relevant log output

_No response_

niklaswolf:
## Changes
closes #1682 

for more details and explanation, have a look in the issue and its comments.

### Checklist


patzick (approved):
🏅 great job!

PR #1688, opened 2021-09-27
fix(default-theme): set negativePattern to default

develth:
## Changes

Set negativePattern to default pattern to have a consistent format, even on negative prices (like promotions)
### Checklist


patzick (approved):
Thank you @develth! I moved your solution to the right file, as it was unused const after refactor :)

Great to see your contribution 🙏

PR #1689, opened 2021-09-28
feat(api-client): expose axios instance from ShopwareApiInstance for advanced use

### LINKED ISSUE #1634: [FEATURE] make axios instance available from ShopwareApiInstance
### Description
It would be nice to have the underlying axios instance available from shopware-6-client, so that it can also be available from `shopwareApiInstance`. 
This is needed to register custom axios interceptors, right now that is impossible as I see it, because you don't have any reference to the axios instance itself.

### Acceptance criteria
- [] Axios instance is available from interface `ShopwareApiInstance`

patzick:
## Changes

closes #1634

exposes Axios instance for advanced use


mkucmus (approved):
there is nothing to hide ;)

PR #1690, opened 2021-09-28
fix(default-theme): address manager switch problems

mkucmus:
## Changes

- solves 500 issues on address change during the checkout


patzick (approved):
🛒

PR #1691, opened 2021-09-28
feat(api-client): change api definitions to public, single getCmsPage endpoint

patzick:
## Changes

- removed `getPage` and `getProductPage` methods, which duplicated logic from `getCmsPage`
- adjusted CMS page interfaces
- changed API service definitions to `public`


mkucmus (approved):
140 files changed is what I want this rainy evening 🌧️ 🌔

Issue #1600, opened 2021-07-19
[BUG] can't access subcategory navigation in SwBottomMenu


PaulSchumacher:
### Project type
- [x] Project (generated by CLI)
- [ ] Contribution (cloned/forked from GH)

### Describe the bug
The SwBottomMenu doesn't let me navigate to subpages. When clicking on the chevron right, only the menu title changes, but not the menu items.

- go to the homepage
- reduce the window size until you see the mobile menu on the bottom of the screen
- Click on the Menu button to open the Menu Modal
- Click on the chevron right on a menu entry that has subcategories
- you should see a swiping transition, the title changes, but not the menu items

### Additional context
- node version: v14.17.3
- shopware-pwa version: 0.91, 0.9.2-canary.1430 
- Shopware 6 detailed version:  6.4.1.0, 6.4.1.2
- system: OSX, CentOS 8

### Screenshots

Starting point
![grafik](https://user-images.githubusercontent.com/15142461/126152691-20a30de1-ef36-4990-8c87-790bbc34a8d5.png)

After clicking on the Service Chevron, the item is was added to the categoryBreadcrumbs array, but the menu isn't changing
![grafik](https://user-images.githubusercontent.com/15142461/126152728-4bc5b54a-ca73-468a-80e8-5a84dd0c0aa0.png)

After clicking on the Service Chevron a few times more
![grafik](https://user-images.githubusercontent.com/15142461/126152748-9a9cca4f-7c26-473e-8108-2145931e8729.png)

patzick:
hey @PaulSchumacher !
Thanks for this report, I cannot reproduce your issue, probably it's been already fixed. Can you confirm that is no longer a problem?

PaulSchumacher:
Hello, I think I have found the bug.

When you create a category in the default language, the name in the second language is empty.
This results in the API-Response containing an empty name, which leads to the broken navigation.

patzick:
Thanks, we'll investigate that!

PR #1695, opened 2021-09-30
fix(default-theme): bottom navigation modal invocation

patzick:
## Changes

While trying to reproduce #1600 I've found that the bottom navi is not opening login window. his PR fixes this issue.


mkucmus (approved):
Good bye, rootContext, our good, old friend 💙 We will miss you 😢

PR #1696, opened 2021-09-30
fix(default-theme): performance issue with css variables duplication

### LINKED ISSUE #1024: [BUG] huge performance issues regarding variables
### Describe the bug
`~@shopware-pwa/default-theme/assets/scss/variables` should only contain SASS variables or mixins, since they are not compiled, it should not contain either styles, classes NOR CSS VARIABLES DEFINITIONS. Or else browsers' will become polluted with lots of variables being re-initialized many many times.

### Proposed solution
Include `@import "~@storefront-ui/shared/styles/variables";`
Include `@import "~@storefront-ui/shared/styles/helpers ";` in `@import "~@shopware-pwa/default-theme/assets/scss/variables";` so we can have mixins from storefront-ui and stuff
Move any variable definition (like this piece of code: ```
#__nuxt {
  --boxed-mode-max-width: 1366px;
  --bottom-navigation-height: 3.35rem;
}```) in main.scss

patzick:
## Changes

closes #1024 
Fixes the perf issue with CSS variables and styles duplication in scoped styles.

In dev mode, it reduced the size of the initial HTML from 494kB -> 291kB
 Before:
image

After:
image

Reduced around 50kB in production build

### DIFF
--- a/packages/default-theme/src/assets/scss/main.scss
+++ b/packages/default-theme/src/assets/scss/main.scss
@@ -3,6 +3,12 @@
 
 // Global styles for theme
 
+// Global variables for theme
+#__nuxt {
+  --boxed-mode-max-width: 1366px;
+  --bottom-navigation-height: 3.35rem;
+}
+
 html {
   overflow-x: hidden;
   height: 100vh;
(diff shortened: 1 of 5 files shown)


mkucmus (commented):
looks and works good 🛒

mkucmus (approved):

PR #1701, opened 2021-10-06
feat(cms): product context for cms elements

patzick:
## Changes

This is a request from the plugins side. When you want to provide a product context for CmsPage and not use default useCms composable you couldn't reach correct product context.

### Requirements
When you use `CmsPage` component and want to pass additional product context then you can use `provide` mechanism inside the `setup` function. Example from `frontend.detail.page`:
```ts
provide("cms-product", product)
```

from then child components can use this product
```ts
const product = inject("cms-product")
```


mkucmus (approved):
sssneaky 🐍

Issue #1353, opened 2021-02-08
[BUG] Double opt-in for registration, newsletter and password recovery


elkmod:
Shopware offers double opt-in functionalities for user and newsletter-registrations as well as a password reset route which works the same way.

In order to support these flows, the application has to "listen" on the following paths and pass the parameters to the respective API-Endpoints mentioned below AND fetch the corresponding configuration (are double-opt ins enabled?) from the API **during build time**:

**Newsletter Confirmation**

PWA-Route: `/newsletter-subscribe?em=<email-hash>&hash=<subscribe-hash>`

API-Route: `POST /newsletter/confirm` => `hash` and `em` body parameters
| Parameter | Description |
|---|---|
|**hash**|Subscribe Hash|
|**em**|Email Hash|

**Registration Confirmation**

PWA-Route: `/registration/confirm?em=<email-hash>&hash=<register-hash>`

API-Route: `POST /account/register-confirm`
| Parameter | Description |
|---|---|
|**hash**|Subscribe Hash|
|**em**|Email Hash|

**Password Recovery**

PWA-Route: `/account/recover/password?hash=<password-hash>`

API-Route: `POST /account/recovery-password-confirm`
| Parameter | Description |
|---|---|
|**hash**|Subscribe Hash|
|**newPassword**|New Password|
|**newPasswordConfirm**|New Password (equality check)|

niklaswolf:
Regarding passwort-recovery: it should be `pwaHost` + current domain ending here: 
https://github.com/vuestorefront/shopware-pwa/blob/master/packages/shopware-6-client/src/services/customerService.ts#L299 the current default is simply wrong :)

PR #1724, opened 2021-11-03
feat: double opt-in registration flow

mkucmus:
## Changes

closes: #1353

### Checklist


packages/default-theme/src/pages/registration/confirm.vue
```diff
@@ -0,0 +1,150 @@
...
+      () => success.value,
+      () => {
+        if (success.value) {
+          refreshUser()
+          redirectTo && redirectUser(redirectTo)
+        }
+      }
+    )
+
+    onDeactivated(() => clearTimeout())
```
mkucmus, line 110: TODO ;)

packages/default-theme/src/pages/checkout.vue
```diff
@@ -164,7 +164,13 @@ export default {
       if (!isFormCorrect) {
         return
       }
-      await register(registrationFormData.value)
+
+      // register user and point /checkout path that may be used in double opt-in (optional usage)
+      await register(
+        Object.assign({}, registrationFormData.value, {
+          redirectTo: encodeURIComponent(PAGE_CHECKOUT),
```
patzick, line 171: we should use this for url, to save domain settings
```suggestion
          redirectTo: routing.getUrl(PAGE_CHECKOUT),
```

`routing` object can be fetched from `getApplicationContext`

packages/default-theme/src/pages/checkout.vue
```diff
@@ -208,6 +214,7 @@ export default {
         // perform a redirection to the external payment gateway
         window.location.href = redirectUrl
       } catch (error) {
+        console.warn("err place order", error)
```
patzick, line 217: `const {devtools} = getApplicationContext()`
```suggestion
        devtools?.warning("[checkout] Error while placing an order", error)
```

packages/default-theme/src/pages/registration/confirm.vue
```diff
@@ -0,0 +1,150 @@
...
+      },
+    } = root
+
+    const success = ref(false)
+    const errors = ref([])
+    const loading = ref(true)
+    const redirectTimer = ref()
+    const redirectUser = (redirectTo) =>
+      setTimeout(() => {
+        router.push(redirectTo)
```
patzick, line 80: is the domain correct here?

patzick (changes requested):
Few suggestions from me

packages/default-theme/src/pages/registration/confirm.vue
mkucmus: we pass here only a path, which is being used in `router.push` once user is confirmed. It's definitely case to be checked 👍🏻

packages/default-theme/src/pages/checkout.vue
mkucmus: thanks, hawk-eye! without routing.getUrl we are always redirected to the default language without any prefix 🙏🏻 now it's fixed. I left encodeURIComponent, path is automatically decoded on vue side 👍🏻

packages/default-theme/src/pages/checkout.vue
mkucmus: done

packages/default-theme/src/pages/registration/confirm.vue
mkucmus: I checked it and router pushes the previously configured path (prepended with current domain' path) as it's passed during registration process. so it's working great!

patzick (approved):
®️ ✅

PR #1746, opened 2021-12-01
fix(nuxt-module): domain configuration now determined properly on route change

### LINKED ISSUE #1745: [Bug] wrong locale for static pages
### What happened?

EDIT: I think this is not only for static pages, but also dynamic pages, see the corresponding PR

I'm on a site setup with different subdomains used for localization. 
I'm on a site with german locale set in the application.
Then I switch to a static page (nuxt page) and the locale suddenly changes to another locale (in my case english).

This happens because of this change here:
https://github.com/vuestorefront/shopware-pwa/commit/01eaf2f9c4d572aa88ab67bacdc1c254011b65d9#diff-76fdfb9d975492f8aceb4679cdd288294f91798f0bd9631358629a764a417be3 

Now not only routes with localization based on the prefixPath get handed the domain config, but all routes, which is fine at first.
But during the routing process, now suddenly `domainConfig` ist set here: 
https://github.com/vuestorefront/shopware-pwa/blob/master/packages/nuxt-module/plugins/domain.js#L144

The comment some lines above states the assumption, that "Only routes with prefixPath have to have the domainConfig added in the meta-object" 
https://github.com/vuestorefront/shopware-pwa/blob/master/packages/nuxt-module/plugins/domain.js#L138 

This of course now breaks, because the assumption isn't valid anymore. Because static-routes always will be the first route from all routes, that matches the new path, the `domainConfig` now will be always the one of the first route in the array matching the path.

This can be solved by also checking, if the new route is a static route, I'll provide a PR asap.

### Steps To Reproduce

_No response_

### Version

nuxt-module

### Environment

```markdown
- Shopware PWA version: 1.0.1
```

### Relevant log output

_No response_

niklaswolf:
## Description

[This recent change](https://github.com/vuestorefront/shopware-pwa/commit/01eaf2f9c4d572aa88ab67bacdc1c254011b65d9#diff-76fdfb9d975492f8aceb4679cdd288294f91798f0bd9631358629a764a417be3) leads to big problems for domains, where localization NOT is based on a prefixPath (e.g. TLD.com and TLD.de). This of course only shows, when there is more than one domain configuration in use.

The change leads to the situation, that all routes have set the corresponding domain-config as meta-information. This is ok, but it breaks the assumption that the implementation in the `domain.js` plugin was based on. The assumption was, that ONLY routes that belong to a domain configuration WITH prefixPath (e.g. TLD.com/de) would have the corresponding domain configuration set as meta information.
The current implementation takes this fact to determine, if the `domainConfig` has to be determined seperately. This has to be done, because of the way the nuxt-router works in combination with the routes-building in the nuxt-module: for each domain, a route is added to the routes-array. The nuxt router picks the first route that has a matching path. 

This works fine, as long as you work with prefixPaths as the prefixPath is part of the route's path (example: `/account` and `/de/account`). The router automatically picks the correct route, because these are actually two seperate routes.
But when you have localization on a host-basis (e.g. TLD.com and TLD.de), the route's paths don't differ anymore, there are actually two routes with the same path, but with other domain configs.
Bildschirmfoto 2021-12-01 um 19 27 18

The router then always picks the first route matching, which does NOT have to be the one with the correct domain config. This not only applies to static routes as I first thought and described in the issue, but also to dynamic routes (see the last two entries in the screenshot).

## Changes
closes #1745.
I have rewritten the process of determining the correct domain configuration based on the new situation, that all routes have a domain configuration set. My implementation's new assumptions are:
1) the routes domain configuration is correct and can be used, when the attached domain configuration has a prefixPath.
2) the routes domain configuration doesn't have to be the right one and has to be determined separately, when the attached domain configuration has NO prefixPath

## Further possible improvements
I noticed a possible improvement, but I didn't want to include it in this PR:
The whole process of determining the correct domainConfig can be skipped, if there is only one domainConfig set in the list of domains.

### Checklist


patzick (approved):
Thank you Niklas! :)

PR #1751, opened 2021-12-06
fix(nuxt-module): optional chaining to hasPrefixPath

### LINKED ISSUE #1750: [Bug] Cannot read property 'pathPrefix' of undefined
### What happened?

Running a fresh install of the latest (1.1.0) results in the following error:

`Cannot read property 'pathPrefix' of undefined` 

### Steps To Reproduce

1. Clean install of shopware-pwa using `npx @shopware-pwa/cli init`
2. Clean install of Shopware production template `v6.4.6.1` => shopware/production
3. fetch domains `npx @shopware-pwa/cli domains` => domains.json is created and set correctly
4. Run `npm run dev` or `yarn dev`
5. Open in browser, open console and the error appears

### Version

nuxt-module

### Environment

```markdown
- OS: macOs
- Node: 14
- Shopware API: 6.4.6.1
- Shopware PWA version: 1.1.0 (assuming we're talking about `@shopware-pwa/nuxt-module`)
```

### Relevant log output

_No response_

mmeester:
closes #1750

### Checklist

### DIFF
--- a/packages/nuxt-module/plugins/domain.js
+++ b/packages/nuxt-module/plugins/domain.js
@@ -140,9 +140,7 @@ export default async ({ app, route, req }, inject) => {
        */
 
       const hasPrefixPath = (domainConfig) => {
-        if (!domainConfig) return false;
-
-        return domainConfig?.pathPrefix !== "/";
+        return domainConfig.pathPrefix !== "/";
       };
 
       // a route can have multiple metadata objects inside - find the one with domainId - it comes from compiled routes (domains nuxt-module)
@@ -154,7 +152,7 @@ export default async ({ app, route, req }, inject) => {
        * The nuxt-router always picks the first route that matches the new path, so for these pages the
        * domainConfig attached to the route doesn't have to be the right one.
        */
-      if (!hasPrefixPath(domainConfig)) {
+      if (domainConfig && !hasPrefixPath(domainConfig)) {
         if (process.client) {
           // During client-side navigation the domain config can be manually loaded through the location origin.
           const currentOrigin = location.origin;
@@ -178,7 +176,7 @@ export default async ({ app, route, req }, inject) => {
       /* Same cause as in the above block:
        * the fromDomainConfig has to be determined separately for routes with a domainConfig without prefixPath
        */
-      if (!hasPrefixPath(fromDomainConfig)) {
+      if (fromDomainConfig && !hasPrefixPath(fromDomainConfig)) {
         if (process.client) {
           // During client-side navigation the domain config can be manually loaded through the location origin,
           // because the origin is still the same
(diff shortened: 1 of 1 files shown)


packages/nuxt-module/plugins/domain.js
```diff
@@ -140,7 +140,7 @@ export default async ({ app, route, req }, inject) => {
        */
 
       const hasPrefixPath = (domainConfig) => {
-        return domainConfig.pathPrefix !== "/";
+        return domainConfig?.pathPrefix !== "/";
```
mkucmus, line 143: If domainConfig is `undefined` the result is `true`. maybe it would be safer if we return false if domainConfig is not set to have false instead of true in this case? WDYT

niklaswolf: I don't know how this can happen, because every route should have the domainConfig now...
Give me some minutes to think this through

niklaswolf: The error can come from two places essentially: 
1) https://github.com/vuestorefront/shopware-pwa/blob/258e6d51193ff2f8eae70b53af62d265fd3e4141/packages/nuxt-module/plugins/domain.js#L149 
2) https://github.com/vuestorefront/shopware-pwa/blob/258e6d51193ff2f8eae70b53af62d265fd3e4141/packages/nuxt-module/plugins/domain.js#L175 

One of those is undefined, so it would be a better fix in my opinion to add a condition to check the existance of `domainConfig` and `fromDomainConfig`:
1) https://github.com/vuestorefront/shopware-pwa/blob/258e6d51193ff2f8eae70b53af62d265fd3e4141/packages/nuxt-module/plugins/domain.js#L157
2) https://github.com/vuestorefront/shopware-pwa/blob/258e6d51193ff2f8eae70b53af62d265fd3e4141/packages/nuxt-module/plugins/domain.js#L181 

So it would be `domainConfig && !hasPrefixPath(domainConfig)` and `fromDomainConfig && !hasPrefixPath(fromDomainConfig)`

Additionally the optionalChaining can be added in `hasPrefixPath`. 
I'd rather not add the condition, because this is not what the function is about: it shouldn't return `false`, when `domainConfig` isn't set, this would be misleading. It should throw an exception instead.

niklaswolf (changes requested):

packages/nuxt-module/plugins/domain.js
mkucmus: I found that the error probably comes from the point 2. (https://github.com/vuestorefront/shopware-pwa/blob/258e6d51193ff2f8eae70b53af62d265fd3e4141/packages/nuxt-module/plugins/domain.js#L175) mentioned by Niklas and it's because that on SSR there is now `from object set and `fromDomainConfig` is always `undefined`. @mmeester could you add another condition for that case?

niklaswolf (approved):

mkucmus (approved):
works fine!

PR #1758, opened 2021-12-10
Add missing field linkNewTab

nicoanders:
## Changes

Adds the field "linkNewTab" to be in sync with the according PHP definition

### Checklist


mkucmus (changes requested):
thanks, feel free to add all the missing fields they are left and not appear here so far 👍🏻 

one little change here: the `linkNewTab` field is also `null` by default.

nicoanders:
Hi @mkucmus,
sure you're right, regarding the default values and I will correct it anyway...
But I had a closer look on the interfaces and now I'm not sure if it's better placed by context here https://github.com/vuestorefront/shopware-pwa/blob/master/packages/commons/interfaces/models/content/navigation/Navigation.ts#L47
Maybe extending from the interface also makes sense, because they are very similar.
What do you think?

mkucmus:
> Hi @mkucmus, sure you're right, regarding the default values and I will correct it anyway... But I had a closer look on the interfaces and now I'm not sure if it's better placed by context here https://github.com/vuestorefront/shopware-pwa/blob/master/packages/commons/interfaces/models/content/navigation/Navigation.ts#L47 Maybe extending from the interface also makes sense, because they are very similar. What do you think?

It's always good to generalize some types and use inheritance to reduce the codesize. Go ahead if you can!

mkucmus (approved):
thanks! we will be developing that on the occasion ;)

PR #1761, opened 2021-12-20
refactor(commons): refactor commons interfaces to be imported directly from the package

patzick:
Cleanup of common interfaces, easier to manage common interfaces import, removed import duplicates as well.


mkucmus (approved):
giphy

PR #1762, opened 2021-12-20
chore: check demo sties availability

mkucmus:
## Changes

adds a cron job via gh action to check demo sites availability


patzick (approved):
Now we know! :)

PR #1766, opened 2021-12-22
fix(default theme):  synchronise tabs after switching between them

patzick:
## Changes

After switching between tabs there was a problem with the synchronisation of the session and card.
For example:
1. go to page
2. open new tam and login
3. go back to page

same with logout

This is the fix for these situations, every time user goes back to the tab we refresh the session and card info to keep that in sync. Additionally, we send interceptor event so you can do additional stuff when it happens using
```
const { on } = useIntercept()

on({
  broadcastKey: 'tab-visible',
  name: "show-console-message",
  handler: () => {
    console.log('Welcome back!')
  },
})
```

image


mkucmus (approved):
magic tricks 🧙🏻

PR #1769, opened 2021-12-22
fix(default-theme): SwTopNaviation visibleCategories null variable

jnuricumbo:
VisibleCategories of SwTopNaviation is sometimes null, for example when switching from checkout to home, by clicking the logo. The issue should be fixed after this change.


patzick (approved):
cool, thank you! :)

PR #1779, opened 2022-01-10
feat(default-theme): test 2e2 with playwright

patzick:
## Changes

Our cypress tests were inconsistent and were having problems with headless running. The Playwright makes testing much more reliable so this is a start to make sure that users can achieve their goals.


mkucmus (approved):
"friendship ended with cypress"
"now playwright is my only friend"
image

PR #1786, opened 2022-01-12
feat(cli): remove devMode question from plugins command

patzick:
## Changes
Removes the question 
```
Do you want to allow dev mode for slots? (Don't use for production!)
```

as it is hard for new members to answer that. Advanced users can still access that by invoking `yarn shopware-pwa plugins --devMode` but our recommendation would be to use VueDevtools to discover plugin slots

image


mkucmus (approved):
less code less troubles ;)

PR #1796, opened 2022-01-19
fix(cli): do not remove @shopware-pwa packages

### LINKED ISSUE #1789: @shopware-pwa/commons package is removed when running init cli command
### What happened?

I import some types from this package as i have a lot of custom composables etc in the project i'm working on. But everytime i run the init i have to reinstall the package manually which is not ideal.

### Steps To Reproduce

1. install @shopware-pwa/commons
2. run shopware-wa init
3. @shopware-pwa/commons is removed from package.json

### Version

CLI

### Environment

```markdown
- OS: macOs
- Node: v14.17.5
- Shopware API: v6.4.2.1
- Shopware PWA version: v1.1.1
```

### Relevant log output

_No response_

patzick:
## Changes

closes #1789

we shouldn't remove @shopware-pwa packages if there're already some of them installed manually. Just update the version of nuxt-module which should be the same as CLI version.

originally proposed solution: #1790 - shoutout to @stigvanbrabant for pointing the problem

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -97,16 +97,6 @@ module.exports = {
       config.dependencies = config.dependencies || {};
       config.devDependencies = config.devDependencies || {};
 
-      // remove all @shopware-pwa packages
-      const shopwarePwaPackageNames = Object.keys({
-        ...config.dependencies,
-        ...config.devDependencies,
-      }).filter((name) => name.includes("@shopware-pwa"));
-      shopwarePwaPackageNames.forEach((packageName) => {
-        delete config.dependencies[packageName];
-        delete config.devDependencies[packageName];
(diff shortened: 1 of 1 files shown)


stigvanbrabant:
@patzick Yeah looks great to me! I don't know what the reasoning was for clearing the packages so i don't really know if it has other side effects tho.

patzick:
it was mostly for backward compatibility because at first we added all of the packages to `package.json` and at some point, we moved them as dependencies of `nuxt-module` so the upgrade is now iterating only one package in the simplest cases.

Nevertheless, I believe it's not needed anymore, especially if that causes problems like yours :)

mkucmus (approved):
you've removed the code that deleting. there is always a bigger fish 🐟

PR #1800, opened 2022-01-27
fix `//` problem when serving from e.g. S3

digitalkaoz:
somehow the urls generated are like `https://domain//pwa/fb3c65966433393af9889753675af9c9.json`
this will fail when serving those from an S3. so we make sure we normalize them before request

### Checklist


packages/cli/src/extensions/plugins-extensions.ts
```diff
@@ -63,7 +63,7 @@ module.exports = (toolbox: GluegunToolbox) => {
         : `${toolbox.normalizeBaseUrl(
             toolbox.inputParameters.shopwareEndpoint
           )}/${config}`;
-    const pluginsConfigResponse = await axios.get(url);
+    const pluginsConfigResponse = await axios.get(url.replace('//', '/'));
```
patzick, line 66: In that case url:
`http://something.com//some/link`
would become:
`http:/something.com/some/link`

which would not be correct

patzick (changes requested):
Thanks for your contribution!
We need a bit of polishing, though.
What would you say to add near the `normalizeBaseUrl` another method like `removeForwardSlashes` with something like:
`.replace(/([^:]\/)\/+/g, "$1")`

this can be reused in more places afterwards

packages/cli/src/extensions/plugins-extensions.ts
digitalkaoz: *welp* for sure you are correct, ill update the PR. it was just a quick edit through the Github UI ;)

digitalkaoz:
@patzick all done. thanks for the hint

patzick (approved):
💙  thanks

PR #1803, opened 2022-02-03
feat(nuxt-module): add parsing of "forwarded" header when checking for valid domains

digitalkaoz:
## Changes

right now only the `X-Forwarded-Host` header is checked when getting the right domain config.
Sadly that wont work in AWS with ALB, there exists the `Forwarded` Header now (Symfony can already deal with that)
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

If present we take this header into account instead of fallback to the internal host (ip) when the `X-Forwarded-Host` header is missing.

### Checklist

### DIFF
--- a/packages/nuxt-module/plugins/domain.js
+++ b/packages/nuxt-module/plugins/domain.js
@@ -13,6 +13,7 @@ const domains = require("sw-plugins/domains");
 const domainsList = Object.values(domains).filter(
   ({ url }) => domainsAllowListArray.includes(url) // TODO: possibly problematic with prefix paths
 );
+import forwardedParse from "forwarded-parse";
 
 // register domains based routing and configuration
 export default async ({ app, route, req }, inject) => {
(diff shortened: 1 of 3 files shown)


mkucmus:
> anyone?

Thanks for you contribution. Now we are about to do some tests on your solution and then will be merged if everything is working well 👍🏻

packages/nuxt-module/plugins/domain.js
```diff
@@ -50,8 +51,19 @@ export default async ({ app, route, req }, inject) => {
...
-          req.headers["x-forwarded-host"] || req.headers.host;
+        // for ALB healtchecks we simply take the first available domain, so we dont generate errors
+        if (req.headers['user-agent'].includes('ELB-HealthChecker') && domainsList.length > 0) {
+          return domainsList[0]
+        }
+
+        let detectedHost = req.headers["x-forwarded-host"] || req.headers.host;
+        // also check the Forwarded header (RFC 7239)
+        if (req.headers["forwarded"]) {
+          const parsedHeader = forwardedParse(req.headers["forwarded"]);
```
mkucmus, line 62: according to the documentation `forwardedParse` function returns an array of objects, not an object. 
so please add a validation and appropriate data extraction, like `parsedHeader?.[0]?.host or filter an array by allowed domain list (utilize `domainsList` variable for that) and then get the host property's value.

mkucmus (changes requested):
Thanks a lot for contributing to the project 💪🏻 

only one comment to do and will be merged for upcoming release 👍🏻

packages/nuxt-module/plugins/domain.js
digitalkaoz: fixed it, as there is only 1 forwarded header, i guess `parsedHeader?.[0]?.host` is perfectly fine

mkucmus: I agree, it's a good starting point of making it more useful in production environment 👍🏻

packages/nuxt-module/plugins/domain.js
```diff
@@ -50,8 +51,22 @@ export default async ({ app, route, req }, inject) => {
         hostname = window.location.hostname;
         pathname = window.location.pathname;
       } else if (!process.static) {
-        const detectedHost =
-          req.headers["x-forwarded-host"] || req.headers.host;
+        // for ALB healtchecks we simply take the first available domain, so we dont generate errors
+        if (
+          req.headers["user-agent"].includes("ELB-HealthChecker") &&
```
mkucmus, line 56: Please use optional chaining in case of user-agent header is somehow empty.
```suggestion
          req.headers["user-agent"]?.includes("ELB-HealthChecker") &&
```

digitalkaoz: done

PR #1833, opened 2022-02-28
chore: bump dependencies

patzick:
## Changes

Quite a bigger upgrade, we had many dependencies in older versions.


mkucmus (approved):
I read those changes with a pleasure :)

PR #1854, opened 2022-03-21
feat(composables): injectable shopware instance

patzick:
## Changes

Leverage Vue mechanism to pass shopware instance


mkucmus (approved):
🤯

PR #1887, opened 2022-06-06
feat(default-theme): Improve SwImage component with srcset

quando1910:
this PR closes https://github.com/vuestorefront/shopware-pwa/issues/1884
## Changes
- add new prop `srcset` for SwImage component.
- update srcset data of SwImage inside CmsElementImage component.

### Checklist
- [x] implement computed property for src set
- [x] find all occurances of SwImage component used in default-theme
- [x] pass src set extracted from available data from the parent (ignore if it's not possible).
- [x] use breakpoints set as default coming with media object (400, 800, 1200?).

### DIFF
--- a/packages/default-theme/src/cms/elements/CmsElementImage.vue
+++ b/packages/default-theme/src/cms/elements/CmsElementImage.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="imgUrl">
+  <div v-if="imagesUrl && imagesUrl.length">
     <SwLink
       v-if="link"
       :link="link"
@@ -7,7 +7,7 @@
       class="cms-element-image__link"
     >
       <SwImage
-        :src="imgUrl"
+        :srcset="imagesUrl"
         :title="title"
         :alt="alt"
         :loading="lazyLoad"
@@ -16,7 +16,7 @@
     </SwLink>
     <SwImage
       v-else
-      :src="imgUrl"
+      :srcset="imagesUrl"
       :title="title"
       :alt="alt"
       :loading="lazyLoad"
@@ -57,8 +57,8 @@ export default {
       return this.content && this.content.data && this.content.data.media
     },
 
-    imgUrl() {
-      return getResizedImage(this.getMedia && this.getMedia.url)
+    imagesUrl() {
+      return this.getMedia?.thumbnails ?? []
     },
 
     lazyLoad() {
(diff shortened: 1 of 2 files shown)


packages/default-theme/src/cms/elements/CmsElementImage.vue
```diff
@@ -1,13 +1,13 @@
 <template>
-  <div v-if="imgUrl">
+  <div v-if="imagesUrl && imagesUrl.length">
```
patzick, line 2: it's a good practice to use optional chaining :)
```suggestion
  <div v-if="imagesUrl?.length">
```

packages/default-theme/src/cms/elements/CmsElementImage.vue
```diff
@@ -57,8 +57,8 @@ export default {
       return this.content && this.content.data && this.content.data.media
     },
 
-    imgUrl() {
-      return getResizedImage(this.getMedia && this.getMedia.url)
+    imagesUrl() {
+      return this.getMedia ? this.getMedia.thumbnails : []
```
patzick, line 61:
```suggestion
      return this.getMedia?.thumbnails ?? []
```

packages/default-theme/src/components/atoms/SwImage.vue
```diff
@@ -11,13 +11,39 @@ export default {
...
   setup(props, { root }) {
     const width = (props.imageWidth && props.imageWidth + "px") || "100%"
     const height = (props.imageHeight && props.imageHeight + "px") || "100%"
     return {
       placeholder: getPlaceholderImage(width, height),
     }
   },
+  computed: {
+    srcsets () {
+      return this.srcset ? this.srcset.map(item => ({
```
patzick, line 30: let's move these computed properties to setup method, example:
```ts

const srcsets = computed(() => props.srcset?.map(item => ({
   src: item.url,
   width: item.width,
   resolution: item.width
  }) ?? []
)

```

patzick (changes requested):
good start! :)
take a closer look at Optional Chaining and look more closely at Vue's Composition API, so we'll have logic inside the `setup` method

packages/default-theme/src/cms/elements/CmsElementImage.vue
quando1910: `<div v-if="imagesUrl?.length">` this code can not work, the terminal always said `Compiled with some errors`. I think the app does not allow us to use optional chaining in `v-if`.

patzick: alright! So let's leave it like that! :)

Issue #1885, opened 2022-06-03
[Bug] Add missing price in horizontal product card view


mkucmus:
### What happened?

![image](https://user-images.githubusercontent.com/5596960/171824684-124bed2b-c5f3-477a-ba7e-e8a23a0a9cf6.png)
In horizontal view on the listing (you can enabe it by clicking the second mode button on right top of the listing)
there is no way to see the prices. Perhaps the Price component is missing, or doesn't have any content to be displayed in the component named SwProductCardHorizontal.vue

SwProductCard.vue component has it implemented correctly as we see;
![image](https://user-images.githubusercontent.com/5596960/171824607-9389dc42-6f99-4850-a828-ec7243db641f.png)

Acceptance criteria:
- [x] logic from SwProductCard component is copied (or even enhanced)
- [x] product prices are displayed properly (depending on product's configuration they can be "variant from" labeled or other way)

### Steps To Reproduce

1. go to any listing page
2. change the view from grid to horizontal
3. browse over the products

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Version

Canary (generated by CLI)

### Which element is affected?

default-theme

### Environment

```markdown
- OS: -
- Node: - 
- Shopware API: latest
- Shopware PWA version: canary
```

PR #1888, opened 2022-06-06
fix(default-theme): Fix missing single price in horizontal view 

rittou:
## Changes

Refers to #1885  

### Checklist

- [x] I have searched the existing issues
- [x] logic from SwProductCard component is copied (or even enhanced)
- [x] product prices are displayed properly (depending on product's configuration they can be "variant from" labeled or other way)


patzick (approved):
All looks great! 👌

PR #1892, opened 2022-06-07
fix: invalid default value for prop "list" in swerrorslist

ghost:
## Changes

- Added the missing factory function to return the default value (the error appears in checkout on order if during the order an error gets thrown)

image

### Checklist


patzick (approved):
thank you 💙

PR #1894, opened 2022-06-08
feat(default-theme): Allow customer to edit their address directly in checkout process

### LINKED ISSUE #1818: [FEATURE] Edit addresses directly in checkout (SwAddressManager)
### Description
Right now a (guest) customer has to add a completely new address when trying to edit a existing one (either from checkout step 1 or from saved customer addresses). This is not a great UX. It should be possible to edit the current address directly. 
It's more or less the same functionality needed as in the standard shopware6 checkout.

It was not possible to implement this, as past shopware versions did not allow to update addresses of guest customers. But this seems to be fixed now (I don't remember which shopware versione exactly), so that there are no more blockers for this functionality.

### Acceptance criteria

- customer has to be able to edit the current address in the second checkout-step.

rittou:
## Changes

Closes #1818 

### Checklist

- [x] Customer has to be able to edit the current address in the checkout-step

### DIFF
--- a/packages/default-theme/src/components/forms/SwAddressManager.vue
+++ b/packages/default-theme/src/components/forms/SwAddressManager.vue
@@ -39,7 +39,7 @@
         class="title desktop-only"
         :level="4"
       />
-      <div v-if="isEditModeOpen" class="address-manager__list">
+      <div v-if="!isEditModeOpen" class="address-manager__list">
         <div
           v-for="address in addresses"
           :key="address.id"
@@ -72,6 +72,7 @@
         v-else
         @success="onAddressSave"
         @cancel="isModalOpen = false"
+        :address=activeAddress
       />
     </SfModal>
   </div>


patzick (changes requested):
Unfortunately, it's not enough, changing shipping/billing address is causing changing the other one as well 
image

Reproduction:
1. not logged in user
2. add the product to the card
3. fill guest registration, single address
4. in checkout summary changing billing address is changing shipping address

rittou:
> 
@patzick In this case, that billing address and shipping address is the same, right?

rittou:
@patzick Did you tick on `Shipping and billing address do not match` checkbox?

patzick:
okay, I checked canary and it's another issue not related to this task :) so all good here!

rittou:
I added more codes for able to add and edit on the checkout page.
Can you take a look at this?

patzick (approved):
👌

PR #1897, opened 2022-06-09
feat(default-theme): Add pagination for placed orders

quando1910:
## Changes
This PR closes https://github.com/vuestorefront/shopware-pwa/issues/379
- update response after fetching orders from API.
- update logic handling pagination for `useCustomerOrders`
- update default limit for `defaultApiParams`
- update UI for `orders` component.

### Checklist
- [x] Pagination component is visible on orders history page
- [x] default pagination params are consistent with the default settings (10 per page)
- [x] consider: do not show pagination if orders collection is less than 11


packages/composables/__tests__/useCustomerOrders.spec.ts
```diff
@@ -65,5 +68,197 @@ describe("Composables - useCustomerOrders", () => {
         expect(orderDetails).toBe(orderResponse);
       });
     });
+
+    describe("getTotal", () => {
```
mkucmus, line 72: great one!

mkucmus (approved):
🚀 that's what pagination should look like 💪🏻

patzick (changes requested):
The code looks very good. Well done. 
Although we have two issues I spotted. The first one is that going to Account -> Order history is sending three post requests. Please investigate why is this happening 
image


The other issue is that we're displaying "0" as number of order history
image


A probable solution for both problems would be using `sharedRef` to share elements between multiple instances of a composable.

quando1910:
@patzick 
this is because we repeat `nuxt-child` 3 times. Each time the route changes, the current page will be rendered 3 times, but only show 1, the remaining pages are hidden. 
I can solve this problem by the below solution. How do you think ?
Screen Shot 2022-06-14 at 11 30 13 AM

patzick (approved):
👌

PR #1898, opened 2022-06-13
fix(default-theme): Adding top navigation container for using out of header

### LINKED ISSUE #1537: [BUG] Using SwTopNavigation out of position
I'm trying to use the SwTopNavigation outside the SfHeader navigation slot. This causes the navigation content to be removed because the Navigation Element offsetTop is greater than 27.

Is this expected behaviour or shouldn't this component be used outside the SfHeader slot on another position?

You can reproduce this bug/behaviour by either adding marginTop to the SwTopNavigation or by placing the SwTopNavigation behind the SfHeader.
1. shopware-pwa o
2. select SwHeader
3. add margin-top to the SwTopNavigation

SwHeader.vue
```
<template>
  <div class="sw-top-navigation" data-cy="main-header">
    <SwCookieBar />

    <SwTopBar />

    <SfHeader
      :title="$t('page.title')"
      class="sw-header sf-header--has-mobile-search"
      :has-mobile-search="false"
      :is-nav-visible="true"
      :is-sticky="false"
    >
      <template #logo>
        <SwLogo />
      </template>
      <template #navigation>
        <div style="margin-top: 10px">
          <SwTopNavigation />
        </div>
      </template>
      <template #search>
        <SwSearchBar />
      </template>
      <template #header-icons>
        <SwHeaderIcons />
      </template>
    </SfHeader>

    <div class="bsn-navigation">
      <SwTopNavigation />
    </div>
  </div>
</template>
```

The Code that causes the bug in the SwTopNavigation
```
navElements.forEach((element) => {
  if (element.offsetTop <= 27) {
    visibleItemsCount += 1
  }
  if (element.offsetTop > 27) {
    unvisibleItemsCount += 1
  }
})

if (unvisibleItemsCount >= 1) {
  // This subtraction makes more space in nav and prevent to move "more category " to the next line.
  this.unwrappedElements = Math.max(0, visibleItemsCount - 2)
} else {
  this.unwrappedElements = visibleItemsCount
}
```

rittou:
## Changes

Closes #1537 

### Checklist

- [x] Able to put SwTopNavigation out of header without error display by using SwTopNavigationContainer component

### DIFF
--- a/packages/default-theme/src/components/forms/SwAddressManager.vue
+++ b/packages/default-theme/src/components/forms/SwAddressManager.vue
@@ -39,7 +39,7 @@
         class="title desktop-only"
         :level="4"
       />
-      <div v-if="isEditModeOpen" class="address-manager__list">
+      <div v-if="!isEditModeOpen" class="address-manager__list">
         <div
           v-for="address in addresses"
           :key="address.id"
@@ -72,6 +72,7 @@
         v-else
         @success="onAddressSave"
         @cancel="isModalOpen = false"
+        :address=activeAddress
       />
     </SfModal>
   </div>
--- a/packages/default-theme/src/components/SwTopNavigationContainer.vue
+++ b/packages/default-theme/src/components/SwTopNavigationContainer.vue
@@ -1,0 +1,18 @@
+<template>
+  <div class="nav-container">
+    <slot></slot>
+  </div>
+</template>
+
+<script>
+export default {
+  
+}
+</script>
+
+<style lang="scss" scoped>
+  .nav-container {
+    z-index: 1;
+    position: relative;
+  }
+</style>


packages/default-theme/src/components/forms/SwAddressManager.vue
```diff
@@ -39,7 +39,7 @@
         class="title desktop-only"
         :level="4"
       />
-      <div v-if="isEditModeOpen" class="address-manager__list">
+      <div v-if="!isEditModeOpen" class="address-manager__list">
```
patzick, line 42: this code is from another PR, shouldn't be here

patzick (changes requested):
Where's the usage of this component?

We can merge that after #1894 or we need to revert changes from that :)

packages/default-theme/src/components/forms/SwAddressManager.vue
rittou: I accidentally checkout from another branch, I will revert it tomorrow.
To the new component, I've just thought if anyone want to use top navigation component, they could use it for avoiding crash.
Any advices here?

patzick: hmm, the problem is more related to some potential bug in Storefront UI component `SfHeader`. Maybe it's not happening anymore after recent library updates?
Did you try to do reproduction steps from the issue? Could you describe the results?

patzick:
@rittou please add info to the issue in order to have instruction. I'll merge this after that so we won't forget :)

rittou:
Yeah, I did leave a comment on this issue for a specific reason and solution for them.

PR #1904, opened 2022-06-29
feat(default-theme): prop for columns count in footer navigation

jissereitsma:
Currently the number of columns is hard-coded, so that - when people use different slot content for this component - the number of columns does not match up. With this PR, people can simply pass in a prop from the parent component (or a wrapper component `src/components/organisms/SwFooterNavigation.vue`) to change the number.

### Checklist


patzick (approved):
👌 thank you @jissereitsma

PR #1909, opened 2022-07-01
fix(default-theme): Use translated name of order.stateMachineState

sr-mothership:
## Changes

In the default theme when the order history in a customer's account is viewed in a different language the order status of each order remains empty, since `.name` is used instead of `.translated.name`.
This PR fixes the component.

### Checklist


mkucmus (approved):
thanks!

PR #1939, opened 2022-08-18
fix(client): update newsletter unsubscribe function

quando1910:
## Changes

Update Request body of when call news letter unsubscribe API


patzick (approved):
👌

PR #1942, opened 2022-08-22
feat(shopware-6-client): add newsletter recipient endpoint

mdanilowicz:
## Changes

Add new newsletter recipient endpoint

### Checklist


packages/shopware-6-client/src/services/customerService.ts
```diff
@@ -385,3 +386,21 @@ export async function setDefaultCustomerPaymentMethod(
...
+/**
+ * Checking if user is newsletter subscriber
+ *
+ * @throws ClientApiError
+ * @public
+ */
+export async function isNewsletterSubscriber(
+  contextInstance: ShopwareApiInstance = defaultInstance
+): Promise<{
+  status: string;
```
mkucmus, line 399: can we expect here some specific values to add it as an enum or union of strings?

mkucmus (approved):
Looks good 👍🏻 only one question to answer and it's ready to go 💯

packages/shopware-6-client/src/services/customerService.ts
mdanilowicz: For now, this endpoint is not described in the documentation, which is why I added string, not enum or union. I think that we can change this type when documentation will be updated and we will be sure about all options

PR #1949, opened 2022-09-13
fix(default-theme): translate text from back button

quando1910:
## Changes

fixes translation in German regarding back button in 404 page.


mkucmus (approved):
no step back 🔙

PR #1955, opened 2022-10-05
feat: add purchase-steps

mdanilowicz:
## Changes

Add purchase steps

### Checklist

- [ ] the [list of features](https://github.com/vuestorefront/shopware-pwa/blob/master/docs/landing/resources/features.md) is updated (if it's a feature and it's relevant)
- [ ] I followed [contributing](https://github.com/vuestorefront/shopware-pwa/blob/master/CONTRIBUTING.md) guidelines

### DIFF
--- a/packages/composables/src/logic/useCartItem.ts
+++ b/packages/composables/src/logic/useCartItem.ts
@@ -38,6 +38,7 @@ export interface IUseCartItem {
   changeItemQuantity: (quantity: number) => Promise<void>;
   removeItem: () => Promise<void>;
   getProductItemSeoUrlData(): Promise<Partial<Product>>;
+  getProductQtySteps: ComputedRef<number[] | null>;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/composables/src/logic/useCartItem.ts
```diff
@@ -91,6 +94,31 @@ export function useCartItem({
 
   const isPromotion = computed(() => cartItem.type === "promotion");
 
+  const getProductQtySteps = computed(() => {
```
mkucmus, line 97: for the `getProductQtySteps()` function, it looks pretty similar like the mentioned helper function. As the interface of `quantityInformation` and `product` are similar for having the same properties (`maxPurchase`, `minPurchase` and `purchaseSteps`) it's tempting to generalize the logic inside packages/helpers/src/product/getProductQtySteps.ts and use it here instead? if the types fit of course 🥲

PR #1974, opened 2023-01-20
fix: typo in interface

TanjaBayer:
Shopware returns extensions not extenstions, seems to be a typo in the interface

## Changes

I did not open an issue for this, just saw that my vscode complained about my testdata taken from a real system, which did not match the provided interaface


patzick (approved):
Great, thank you! :)

PR #1979, opened 2023-02-03
fix(default-theme): carousel not refresh when cross selling change tab

quando1910:
## Changes
- Change SwProductCarousel & SwProductCrossSells component.

### Checklist

- [x] change tab will refresh SwProductCarousel


patzick (approved):
👌  Thanks Quan!

PR #1986, opened 2023-03-30
chore: update vuevalidate dependencies

quando1910:
## Issue
- After yarn start, can not run `test-project` because of the version changes of @vuelidate.
## Changes
-  Change to static version of `@vuelidate/core` & `@vuelidate/validators`


packages/cli/src/templates/project-template/package.json
```diff
@@ -21,8 +21,8 @@
     "@shopware-pwa/nuxt-module": "latest",
     "@storefront-ui/vue": "0.13.2",
     "@vue/test-utils": "^2.0.0",
-    "@vuelidate/core": "^2.0.0-alpha.41",
-    "@vuelidate/validators": "^2.0.0-alpha.29",
+    "@vuelidate/core": "2.0.0",
```
patzick, line 24: The newest one is `2.0.2` maybe it's worth doing this?

patzick (changes requested):

packages/cli/src/templates/project-template/package.json
quando1910: No, 2.0.2 is the cause of this bug 
Screen Shot 2023-03-31 at 10 24 03 AM

patzick: ok so let's leave that for now :)

patzick (approved):

